PR #2, opened 2022-07-05
fixing for RO-3041

zkzfng:
Hi Jose,

I fixed the issue and created a new branch. Could you review it and push it to the production?

Thanks,

Zukang


josemduarte (approved):
Code looks good to me. But it'd be nice to do a review of the images after this change. See my comment in RO-3041.

PR #4, opened 2023-02-28
Stashed Cache Resource Update: standardizing workflow

aliciaaevans:
Updates to ProteinTargetSequenceExecutionWorkflow to standardize the way it is called with the other similar workflows


rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
```diff
@@ -31,16 +31,30 @@
...
+            configPath (str, optional): path to configuration file (default: exdb-config-example.yml)
+            configName (str, optional): configuration section name (default: site_info_configuration)
+            cachePath (str, optional):  path to cache directory (default: '.')
+            stashRemotePrefix (str, optional): file name prefix (channel) applied to remote stash file artifacts (default: None)
+        """
+        configPath = kwargs.get("configPath", "exdb-config-example.yml")
+        self.__configName = kwargs.get("configName", "site_info_configuration")
+        mockTopPath = kwargs.get("mockTopPath", None)
+        self.__cfgOb = ConfigUtil(configPath=configPath, defaultSectionName=self.__configName, mockTopPath=mockTopPath)
+        #
```
piehld, line 47: So it looks like you're removing the use of the `HERE` parameter. I thought using that was consistent throughout other Workflow classes, or is it not?

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
```diff
@@ -31,16 +31,30 @@
...
+        self.__cachePath = kwargs.get("cachePath", ".")
```
piehld, line 48: Usually, we use a directory called "CACHE", rather than putting all the files in the working directory. Is there a reason you changed that?

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
```diff
@@ -182,7 +205,7 @@ def fullWorkflow():
     ptsWf = ProteinTargetSequenceExecutionWorkflow()
     ok = True
     ok = ptsWf.cacheTaxonomy()
-    ok = ptsWf.fetchUniProtTaxonomy()
+    ok = ptsWf.updateUniProtTaxonomy()
```
piehld, line 208: Why not keep `fetchUniProtTaxonomy()`? It looks like the `updateUniProtTaxonomy()` method takes about an hour to run. Is it necessary to build it from scratch every time?

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
```diff
@@ -114,7 +128,8 @@ def exportFasta(self):
         ok = False
         try:
             ptsW = ProteinTargetSequenceWorkflow(self.__cfgOb, self.__cachePath)
-            ok = ptsW.exportTargetsFasta(useCache=True, addTaxonomy=True, reloadPharos=True, fromDbPharos=True, resourceNameList=["sabdab", "card", "drugbank", "chembl", "pharos"])
+            ok = ptsW.exportTargetsFasta(useCache=True, addTaxonomy=True, reloadPharos=True, fromDbPharos=True,
+                                         resourceNameList=["sabdab", "card", "drugbank", "chembl", "pharos"])
```
piehld, line 132: I'm surprised pylint let you do that. Usually the formatting requires that for multiline parentheses, it should look like:
```python
ok = ptsW.exportTargetsFasta(
    useCache=True, addTaxonomy=True, reloadPharos=True, fromDbPharos=True, resourceNameList=["sabdab", "card", "drugbank", "chembl", "pharos"]
)
```

For consistency purposes with other projects, do you mind reformatting that?

piehld: (as well as the other methods here?)

Another thing I thought it usually requires is that when you give multiple arguments that extend beyond the 185 character limit, you must split them into a line per argument—not, say, 5 arguments on line one and 5 arguments on a second line. I know it's yelled at me for that before, so I'm surprised it's passing for you now.

piehld (commented):
Thanks for the changes. I made a couple questions and comments throughout the code.

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
aliciaaevans: I don't think so. I did a quick search and it's in `PubChemExecutionWorkflow` but not most of the others, e.g. those in https://github.com/rcsb/py-rcsb_exdb/tree/master/rcsb/exdb/wf or https://github.com/rcsb/py-rcsb_workflow/tree/master/rcsb/workflow/chem .

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
aliciaaevans: Hmm, I believe I copied this from another workflow and didn't notice that. Maybe I picked a bad example. I'll take a look.

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
aliciaaevans: The previous way, that file was never getting updated in BuildLocker. updateUniProtTaxonomy is not called from anywhere else that I could find. That was one of the files that had a modified date of before October 2021 in BuildLocker. I'm open to suggestions on a better way to handle this, if you think it's going to be a problem.

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
aliciaaevans: Hmm, maybe I don't have pylint configured correctly on my IDE, I'll take a look.

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
piehld: Hm OK. I still feel it would be good to keep that as the default though (i.e., if you don't specify a working path, then use the directory of the file a.k.a. "HERE"). You may need to add a new parameter `workPath` for that (which you'll find all over in other packages).

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
piehld: Yeah it would be good to address. I just noticed you also made that change in your last merge to `rcsb.utils.dictionary` (sorry I didn't catch that earlier). Would be good to keep that consistent too.

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
aliciaaevans: Can you give me an example of a file that you think has a good configuration to use as a guide? I made the changes to match those files I linked to above, so we should probably update those other ones too if this isn't right.

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
aliciaaevans: Fixed the sloppy wrapping and reverted the unnecessary wrapping. Turns out my IDE guide was also set to 120, not 185.

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
aliciaaevans: As we discussed, `fetchUniProtTaxonomy()` does not update the stash and there is no incremental update, unfortunately. So, keeping `updateUniProtTaxonomy()`.

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
aliciaaevans: Just pushed these suggested updates. Thanks!

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
aliciaaevans: I'll review the other PRs where I made similar changes.

piehld (approved):
Thanks @aliciaaevans! Changes look good.

rcsb/workflow/targets/ProteinTargetSequenceExecutionWorkflow.py
piehld: Sounds great, thank you!

PR #6, opened 2023-03-14
Feature: V0.36 Generate CARD annotations instead of features

piehld:
This change is needed in order for the resource-building workflow to begin generating the new CARD-annotations data file and putting it on BL and Git Stash.

Related Jira ticket: https://rcsbpdb.atlassian.net/browse/RO-3781


aliciaaevans (approved):
Looks good. Good thinking on the extra logging, I think that will be helpful.

PR #8, opened 2023-05-05
Support restoring database files from stash for Pharos-targets

aliciaaevans:
Due to the extra infrastructure resources and time needed to pull the Pharos data in mySQL dump format from http://juniper.health.unm.edu/tcrd/, coupled with infrequent updates (last one was September 2022), we don't need to do this during the weekly update.

The weekly-update-workflow will be able to pass in False and the files will be restored from BuildLocker stash. All related data from Pharos providers will be built use those tdd files from the local CACHE.

The update of the stashed tdd files from the mySQL dump can be still be done ad-hoc when TCRD is updated.

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -27,4 +27,4 @@
  3-Mar-2023 - V0.35 Standardize args in ProteinTargetSequenceExecutionWorkflow
 14-Mar-2023 - V0.36 Generate CARD annotations instead of features during ProteinTargetSequenceExecutionWorkflow
 21-Mar-2023 - V0.37 Allow backing up Pharos-targets to stash
- 5-May-2023 - V0.38 Add fromDbPharos parameter to ProteinTargetSequenceExecutionWorkflow.exportFasta()
+ 5-May-2023 - V0.38 Add fromDbPharos and reloadPharos parameters to ProteinTargetSequenceExecutionWorkflow.exportFasta()
--- a/rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
+++ b/rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
@@ -11,7 +11,7 @@
 #   3-Mar-2023 Fix error for missing taxonPath
 #  14-Mar-2023 Generate CARD annotations instead of CARD features
 #  21-Mar-2023 Allow backing up Pharos-targets to stash
-#   5-May-2023 Restore from stash if in fromDbPharos and reloadPharos parameters are False
+#   5-May-2023 Restore from stash if fromDbPharos and reloadPharos parameters are False
 ##
 __docformat__ = "google en"
 __author__ = "John Westbrook"


rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
```diff
@@ -474,7 +479,7 @@ def __buildCofactorData(self, referenceResourceName, resourceName, backup=False,
                 aP = PharosTargetCofactorProvider(cachePath=self.__cachePath, useCache=True, useStash=True, useGit=True)
                 ok = aP.buildCofactorList(resultPath, crmpObj=crmpObj, lnmpObj=lnmpObj, maxActivity=maxActivity)
                 if backup:
-                    okB = aP.backup(self.__cfgOb, self.__configName, remotePrefix=remotePrefix)
+                    okB = aP.backup(self.__cfgOb, self.__configName, remotePrefix=remotePrefix, useGit=True)
```
aliciaaevans, line 482: This was the only file that wasn't updating in GitHub without calling DictMethodResourceProvider.syncCache. We may decide to stop updating to git, but that can be a separate decision.

rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
```diff
@@ -181,6 +182,10 @@ def __exportTargetsFasta(self, resourceName, useCache=True, addTaxonomy=False, r
                     if backupPharos:
                         okB = ptP.backup(self.__cfgOb, self.__configName, remotePrefix=remotePrefix, useStash=True, useGit=True)
                         logger.info("%r targets backup status (%r)", resourceName, okB)
+                elif not (reloadPharos or fromDbPharos):
```
piehld, line 185: So basically the logic is:
- If `ptP.testCache()` succeeds, backup the Pharos data
- elif `ptP.testCache()` fails, restore from stash

Am I following that right?

piehld (approved):
Looks good, just a few comments/questions made throughout the PR.

And some other comments:
> The update of the stashed tdd files from the mySQL dump can be still be done ad-hoc when TCRD is updated.

How should we keep track of when those data are updated? Should we just set a calendar reminder every month or so to check?

Also, did that Azure error from this past weekend fix itself?

rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
aliciaaevans: > Just to clarify, you mean this file ([Pharos-cofactors](https://github.com/rcsb/py-rcsb_exdb_assets_stash/tree/development/stash/Pharos-cofactors))? And this change will cause it to be updated every week on GitHub (whenever the new Luigi cache-rebuilding task runs)?
Yes.

> Will all other files be updated on [py-rcsb_exdb_assets_stash](https://github.com/rcsb/py-rcsb_exdb_assets_stash/tree/development/stash) every week too?
Yes. As it stands now, anyway, that was the behavior for the update scripts. `DictMethodResourceProvider.syncCache()` was actually creating duplicate commits for the others that it was syncing, so it makes sense to not call that and just take care of it here.

HISTORY.txt
```diff
@@ -26,4 +26,5 @@
  9-Jan-2023 - V0.34 Configuration changes to support tox 4
  3-Mar-2023 - V0.35 Standardize args in ProteinTargetSequenceExecutionWorkflow
 14-Mar-2023 - V0.36 Generate CARD annotations instead of features during ProteinTargetSequenceExecutionWorkflow
-21-Mar-2023 - V0.37 Allow backing up Pharos-targets to stash
\ No newline at end of file
+21-Mar-2023 - V0.37 Allow backing up Pharos-targets to stash
+ 5-May-2023 - V0.38 Add fromDbPharos parameter to ProteinTargetSequenceExecutionWorkflow.exportFasta()
```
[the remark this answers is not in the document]
aliciaaevans: Thanks, will do.

rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
aliciaaevans: Yeah, if reloadPharos or fromDbPharos are both false, nothing really gets done inside `PharosTargetProvider`. Although, now that I'm thinking of it, perhaps just an `else` would be better. Do you think we should restore the backup if something went wrong in loading the data from the SQL dump?

piehld: Do you mean if something goes wrong during the loading of data from the SQL dump obtained from Pharos' site directly? If so, I think in that case it's best for it to totally fail, because we would only be doing that if we want to rebuild the cache files, right?

aliciaaevans:
> How should we keep track of when those data are updated? Should we just set a calendar reminder every month or so to check?

Good question. Perhaps something could be set up to read the version from or updated date of http://juniper.health.unm.edu/tcrd/download/latest.README and send a notification, but I'm not sure how reliable that would be.

> Also, did that Azure error from this past weekend fix itself?

Yes, the rerun succeeded.

rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
aliciaaevans: Good point

piehld:
> Good question. Perhaps something could be set up to read the version from or updated date of http://juniper.health.unm.edu/tcrd/download/latest.README and send a notification, but I'm not sure how reliable that would be.

That would definitely be the best option, but would obviously take a lot more work to do. Another similar approach could be to incorporate the version info into one of the stashed cache files, and then whenever it tries to rebuild from scratch it compares the version in our stash with that from Pharos. Then, if there's a newer version proceeds to rebuild from scratch; else, it restores the data from stash.

In any case though, since we're not yet sure on the future of Pharos data integration, I think a calendar reminder every month or two would be good enough for now. (Obviously feel free to invite me to it too).

For record's sake, Pharos is currently at version:
```
TCRD v6.13.4
Release Date: 20220908
```

rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
piehld: OK cool, thanks for confirming and explaining.

PR #9, opened 2023-05-22
Add retries to tox task for MMseqs2 download

piehld:


aliciaaevans (approved):
Looks good, and I can see that the retry worked as expected in Azure.

PR #10, opened 2023-06-01
Stop backing up stash files to GitHub (RO-3813)

aliciaaevans:
Change the stashed file update workflows to not use GitHub for backups. The restore of files will still use the sftp server (Buildlocker) with restore from GitHub as a backup.

I preferred to set the defaults for backing up to GitHub to False because I think we may decide to discontinue using GitHub for this completely in favor of something that does a better job of handling large files.

If we want to do an ad-hoc update, `DictMethodResourceCacheWorkflow.syncResourceCache()` can be used to restore files from Buildlocker and sync to GitHub.


rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
```diff
@@ -472,21 +473,21 @@ def __buildCofactorData(self, referenceResourceName, resourceName, backup=False,
...
-                    okB = aP.backup(self.__cfgOb, self.__configName, remotePrefix=remotePrefix, useStash=True, useGit=True)
+                    okB = aP.backup(self.__cfgOb, self.__configName, remotePrefix=remotePrefix, useStash=True, useGit=False)
                     logger.info("%r cofactor backup status (%r)", resourceName, okB)
 
             elif resourceName == "pharos":
                 aP = PharosTargetCofactorProvider(cachePath=self.__cachePath, useCache=True, useStash=True, useGit=True)
                 ok = aP.buildCofactorList(resultPath, crmpObj=crmpObj, lnmpObj=lnmpObj, maxActivity=maxActivity)
                 if backup:
-                    okB = aP.backup(self.__cfgOb, self.__configName, remotePrefix=remotePrefix, useGit=True)
+                    okB = aP.backup(self.__cfgOb, self.__configName, remotePrefix=remotePrefix, useStash=True, useGit=False)
```
piehld, line 483: Interesting that this was missing before. Were there any unintended consequences of this, or was `useStash` already defaulting to `True` anyway?

piehld (approved):
LGTM, thanks! Just one question out of curiosity below.

rcsb/workflow/targets/ProteinTargetSequenceWorkflow.py
aliciaaevans: It does default to true, but I wanted to make it explicit to match the other calls and also for ease of understanding.

PR #13, opened 2023-08-08
V0.42 Don't overwrite Buildlocker files if no data found

piehld:
- Integrating PR from @aliciaaevans: https://github.com/rcsb/py-rcsb_workflow/pull/12

> We've seen that sometimes when there is a failure to get data from an external source, the resource file is still created and stashed, overwriting existing data which could be used until the external source is available. (IMGT https issue, for example.)
> 
> This change is to make sure the Buildlocker file is only updated if the returned ok is True.
> 
> Note, there is a dependency on each provider to report its status correctly. This PR is meant to make sure ok is not True if providers that the current provider relies on does not have data: https://github.com/rcsb/py-rcsb_utils_targets/pull/18
> 

- Also reduce resource footprint for Azure tests


piehld:
Status update:
- The changes made to mitigate the heavy resource demand in Azure all appear to work OK.  This basically involved the addition of the smaller test files in `test-data/chem_comp/`.
- With regards to the other part of this ticket, there are some very complicated issues with the current state of the `ProteinTargetSequenceWorkflow.py` code, which will require elaborate fixing in both this repository and other repositories (mainly `py-rcsb_utils_targets`, but also `chemref` and maybe others). The specific issues are:
  - We cannot rely on adding `providerX.testCache()` calls to check if data is available to export because it can return `False` even if the provider has been initialized and used, if the `.reload()` method isn't called again first.
  - Further, we cannot rely on adding `.reload()` calls immediately before the `.testCache()` calls either, because this would cause the provider to reload the existing data in CACHE, thus overwriting the newly constructed data stored in memory. So, doing this would lead to the writing out of the same data that you started with (since the new in-memory data hasn't been exported and pushed to stash yet).
  - The `testCache()` methods for each provider differ in their requirements, several of which contain bad logic (e.g., see https://github.com/rcsb/py-rcsb_utils_targets/pull/18; however, note that even the proposed fix there would lead to issues in the corresponding test cases, which for some reason expect `testCache` to succeed even if no data is loaded by the provider to begin with)
  - The `__reload()` methods for each provider differ as well, as many providers only try to load data from a local file, and do nothing if it's not present. Normally, we'd expect the `__reload()` method to first try to load from file, and if not present then fetch the external data and rebuild the resource.

piehld:
Just a follow-up: I ended up going with a combination of the two approaches I proposed in my last comment. Specifically:
- Added `.reload()` steps prior to the `testCache()` calls in the workflow code here to ensure the latest cache data has been reloaded into memory
- Updated many of the resource providers in [py-rcsb_utils_targets](https://github.com/rcsb/py-rcsb_utils_targets) (https://github.com/rcsb/py-rcsb_utils_targets/pull/20 and https://github.com/rcsb/py-rcsb_utils_targets/pull/21) and [py-rcsb_exdb](https://github.com/rcsb/py-rcsb_exdb) (https://github.com/rcsb/py-rcsb_exdb/pull/15) to properly test their cache as well as the cache of their dependencies

This second task will be an ongoing process, in case some providers have been overlooked. But this is a start and has been tested to work successfully.

PR #19, opened 2024-12-11
Mt adding imgs workflow

trumbullm:
Added new imgs workflow and its cli.


rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,154 @@
...
+import datetime
+import random
+from pathlib import Path
+import logging
+import subprocess
+
+logger = logging.getLogger(__name__)
+
+
+def get_pdb_list(args: dict) -> list:
```
josemduarte, line 14: In all functions in this file: my style preference would be to use individual type parameters (e.g. `get_pdb_list(pdb_gz_path: str, update_all_images: bool)` instead of the `args` dictionary. Passing the args dictionary around requires to have the argument names hard-coded throughout the code, introducing a tight coupling between the CLI interface and the functions here.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,154 @@
...
+            '/usr/bin/xvfb-run',
+            '-a',
+            '-s', '-ac -screen 0 1280x1024x24',
+            args["molrender_exe"],
+            'all',
+            bcif_file_path,
+            out_path,
+            '--height', '500',
+            '--width', '500',
+            '--format', 'jpeg'
```
josemduarte, line 132: Since this is now a more general CLI outside the weekly update codebase I think it'd be important to extract the different parameters as CLI parameters, e.g. image size, xvfb executable path (possibly an optional parameter, because some environments may not require xvfb), etc.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,154 @@
...
+            ]
+        try:
+            result = subprocess.run(cmd, capture_output=True, text=True, check=True)
+            logger.info("Command was successful!")
+            logger.info(result.stdout)
+        except subprocess.CalledProcessError as e:
+            msg = f"Command failed with exit code {e.returncode} \n Error output: {e.stderr}"
+            logging.exception(msg)
```
josemduarte, line 140: This logs a message but doesn't thrown an exception, right? That wouldn't signal to the caller (the workflow engine) that the task failed.

josemduarte (changes requested):
Thanks! A few comments from me below.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
trumbullm: @josemduarte I originally passed them into variables but Douglas' ruff linter suggested I use something else. The reason being it only recommends passing <5 variables into a function for readability. I can change the linters settings of course but is there another potential solution you could recommend?

trumbullm: Actually, I see you point is specifically about `get_pdb_list()` which does not have this problem since it doesn't require as many variables. What do you think of me changing `get_pdb_list()` and `get_csm_lsit()` to accepting variables and leaving `images_gen_lists()` and `images_gen_jpgs()` to accept the dict?

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
trumbullm: I've added `raise` here as a potential solution. From searching around it seems that is appropriate because it will raise the `CalledProcessError`. If you'd prefer a specific other exception to be raised let me know.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
trumbullm: I just added them!

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
piehld, line 1: Just a general request: Could you make this file a class, called `PdbCsmImageWorkflow`? (E.g., following the similar structure as https://github.com/rcsb/py-rcsb_exdb/blob/master/rcsb/exdb/wf/ExDbWorkflow.py#L43)

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,157 @@
+"""Images workflow functions that actually do the work."""
+
```
piehld, line 2: Could you please add the similar header information as provided in https://github.com/rcsb/py-rcsb_exdb/blob/master/rcsb/exdb/wf/ExDbWorkflow.py#L1-L16 ?

(Obviously tailor it for you, and the Update section would just have one line saying something like, "Created class")

rcsb/workflow/cli/__init__.py
```diff
@@ -0,0 +1,5 @@
+__docformat__ = "google en"
+__author__ = "John Westbrook"
+__email__ = "john.westbrook@rcsb.org"
+__license__ = "Apache 2.0"
+__version__ = "0.48"
```
piehld, line 5: Nice catch on updating the version number!

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,157 @@
...
+                t1 = Path.stat(bcif_file).st_mtime
+                t2 = metadata["datetime"].timestamp()
+                if t1 < t2:
+                    model_list.append(f"{model_id} {model_path} computational")
+            else:
+                model_list.append(f"{model_id} {model_path} computational")
+    return model_list
+
+
+def images_gen_lists(args: dict) -> None:
```
piehld, line 73: If you need to provide a large dict of arguments (instead of specifying them each explicitly), I would recommend using `**kwargs` instead of an argument called `args`. E.g., see usage at https://github.com/rcsb/py-rcsb_exdb/blob/master/rcsb/exdb/wf/ExDbWorkflow.py#L76 (in which you then grab the arguments you need using `kwargs.get(...)`, which lets you also specify default values).

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,157 @@
...
+    random.shuffle(full_id_list)
+
+    steps = int(len(full_id_list) / int(args["num_workers"]) )
+    for i in range(0, len(full_id_list), steps):
+        Path(args["id_list_path"]).mkdir(parents=True, exist_ok=True)
+        with Path.open(args["id_list_path"] + str(int(i/steps)), "w") as f:
+            for line in full_id_list[i : i + steps]:
+                f.write(line + "\n")
+
+def images_gen_jpgs(args: dict) -> None:
```
piehld, line 94: Same comment here regarding `**kwargs`

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,157 @@
...
+            args["molrender_exe"],
+            'all',
+            bcif_file_path,
+            out_path,
+            '--height', args['jpg_height'],
+            '--width', args['jpg_width'],
+            '--format', args['jpg_format'],
+            ]
+        if args['jpg_additional_cmds'] is None:
+            cmd = [*cmd, args['jpg_additional_cmds']]
```
piehld, line 135: We're going to have to think about how we'd enable testing of this (i.e., how to install the molrender software into Azure, or GitHub Actions when we get there).

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,157 @@
...
+            args['jpg_xvfb_executable'],
+            '-a',
+            '-s', f'-ac -screen 0 {args['jpg_screen']}',
+            args["molrender_exe"],
+            'all',
+            bcif_file_path,
+            out_path,
+            '--height', args['jpg_height'],
+            '--width', args['jpg_width'],
+            '--format', args['jpg_format'],
```
piehld, line 132: Here and elsewhere, for our ExDB python code, we generally standardize using double-quotes instead of single quotes, when possible.

piehld (changes requested):
Thanks @trumbullm! Overall looks very good. I had a few suggestions that I left comments on. As I mentioned in person, one general thing that I'll leave in the air for now is that all of our Python ExDB codebases use camel casing instead of snake casing, so it would be ideal if your code could follow that standard too...(I think there might be VS code extensions that would automate that task for you, actually...).

Regarding the Azure formatting errors, we'll talk about those over zoom shortly.

piehld:
@trumbullm Oh also, one other remaining thing to do is add a test file. You can follow the example here for a starting point: https://github.com/rcsb/py-rcsb_exdb/blob/master/rcsb/exdb/tests/testExDbWorkflow.py

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
trumbullm: @piehld I set the default values in the arg parse. Should I instead set them in .get() ?

piehld: I don't think it hurts to have redundancy, so I would suggest leaving them in the arg parse but doing so again here.

trumbullm:
@josemduarte @piehld Maybe we wait until January to merge this?

piehld:
@trumbullm That's what my preference would be. If you were to merge it, it shouldn't affect production at all, but still at that point I don't see the urgency in merging this until we're ready to use it. We also still need the tests as I mentioned.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
josemduarte: As advice by @piehld let's follow the current convention of using `kwargs`

trumbullm: I've changed each function to `**kwargs` now so I'll close this thread.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
trumbullm: So this task is particularly weird... the rcsb-wuw task will build a shell command to be sumitted to Nersc... Done. The shell command is a script that builds a new shell command for each item in a file then runs that command. Why not just combine these shell-building-commands into one? Because the second one needs to open and read the file containing the pdbs and csm to do. Currently, this file is located at Nersc, so they can't be combined. The way around this would be to submit the task with the specific pdb or csm to do instead of writing a list of them to a file. Not writing to a file has the problem of rerunning things if they fail. Yes we can with airflow, but things can still break easier this way. This file version is much simpler as well.

trumbullm:
I'm still developing it a lot and each time I push it runs tests each time. I'll just resubmit a merge request when it's ready.

PR #20, opened 2025-01-15
Mt adding imgs workflow

trumbullm:
Imgs workflow includes an idList generator (has tests) to build lists of pdb and csm ids that need images and a jpg generator (no tests because including node is excessive) which produces the actual jpgs. It uses the standard shell calling scheme as other rcsb packages.


rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,86 @@
...
+
+logging.basicConfig(level=logging.INFO, format="%(asctime)s [%(levelname)s]-%(module)s.%(funcName)s: %(message)s")
+logger = logging.getLogger()
+
+
+class PdbCsmImageWorkflow:
+
+    def imagesGenJpgs(self, **kwargs: dict) -> None:
+        """Generate jpgs for given pdb/csm list."""
+        idListName = kwargs.get("idListName")
```
piehld, line 31: Along with this kwarg-extracted variable, I would extract all possible kwargs up here and then reference the variables down below. That makes it clearer to the user (i.e., us) as to which keywords arguments are actually possible/available.

Also, when you extract them, be sure to assign a default value, such as  
```python

idListName = kwargs.get("idListName", None)
```

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,86 @@
...
+                kwargs.get("jpgXvfbExecutable"),
+                "-a",
+                "-s", f"-ac -screen 0 {kwargs.get('jpgScreen')}",
+                kwargs.get("molrenderExe"),
+                kwargs.get("jpgRender"),
+                bcifFilePath,
+                outPath,
+                "--height", str(kwargs.get("jpgHeight")),
+                "--width", str(kwargs.get("jpgWidth")),
+                "--format", kwargs.get("jpgFormat"),
```
piehld, line 66: see my earlier comment about extracting kwargs above.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,86 @@
...
+            ]
+            if kwargs.get("jpgAdditionalCmds"):
+                cmd = [*cmd, kwargs.get("jpgAdditionalCmds")]
```
piehld, line 69: and here

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,86 @@
...
+                    result = subprocess.run(cmd, capture_output=True, text=True, check=True)
+                    logger.info(result.stdout)
+                except subprocess.CalledProcessError as e:
+                    logger.error("Error: %s", e)
+                    logger.error("Stderr: %s", e.stderr)
+
+                # check result
+                outJpgFile = os.path.join(outPath, fileId + "_model-1.jpeg")
+
+                if not (Path(outJpgFile).is_file() and Path(outJpgFile).stat().st_size > 0):
```
piehld, line 83: Can you follow my other suggestion here as well, for defining the Path() object first, then referencing it. (Instead of creating the object twice in a row)

piehld (changes requested):
Thanks @trumbullm! Please see my comments and suggestions below.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,86 @@
...
+        idListName = kwargs.get("idListName")
+        idListFile = os.path.join(kwargs.get("idListPath"), idListName)
+        logger.info('using id file %s', idListFile)
+
+        if not (Path(idListFile).is_file() and Path(idListFile).stat().st_size > 0):
+            logger.warning('Missing idList file %s', idListFile)
+            return
+
+        with Path(idListFile).open("r", encoding="utf-8") as file:
+            idList = [line.rstrip("\n") for line in file]
```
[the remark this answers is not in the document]
piehld: @trumbullm Here's an example: https://github.com/rcsb/py-rcsb_db/blob/master/rcsb/db/wf/RepoLoadWorkflow.py#L118-L120

trumbullm:
@piehld I think I addressed each comment you made.

rcsb/workflow/cli/ImgExec.py
```diff
@@ -0,0 +1,67 @@
...
+            "genLists",
+            "genJpgs",
+        ],
+    )
+
+    parser.add_argument("--pdbBaseDir")
+    parser.add_argument("--csmBaseDir")
+
+    parser.add_argument("--idListPath")
+    parser.add_argument("--idListName")
```
josemduarte, line 37: How about making these 2 a single parameter: something like `idListFilePath`

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,99 @@
...
+            if jpgAdditionalCmds:
+                cmd = [*cmd, jpgAdditionalCmds]
+
+            fileObj = Path(bcifFilePath)
+            if fileObj.is_file() and fileObj.stat().st_size > 0:
+                # logger.info('Running %s', ' '.join(cmd))
+                try:
+                    result = subprocess.run(cmd, capture_output=True, text=True, check=True)
+                    logger.info(result.stdout)
+                except subprocess.CalledProcessError as e:
```
josemduarte, line 89: This catches the exception producing an error in logs. But it would not signal the caller (Airflow) that it is a failure. For that you would need to rethrow the exception.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,99 @@
...
+                    logger.error("Error: %s", e)
+                    logger.error("Stderr: %s", e.stderr)
+
+                # check result
+                outJpgFile = os.path.join(outPath, fileId + "_model-1.jpeg")
+                fileObj = Path(outJpgFile)
+                if not (fileObj.is_file() and fileObj.stat().st_size > 0):
+                    logger.error("No image file: %s.", outJpgFile)
```
josemduarte, line 97: This seems to be a serious enough error that requires an exception thrown so that the workflow knows about the problem.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,99 @@
...
+                # logger.info('Running %s', ' '.join(cmd))
+                try:
+                    result = subprocess.run(cmd, capture_output=True, text=True, check=True)
+                    logger.info(result.stdout)
+                except subprocess.CalledProcessError as e:
+                    logger.error("Error: %s", e)
+                    logger.error("Stderr: %s", e.stderr)
+
+                # check result
+                outJpgFile = os.path.join(outPath, fileId + "_model-1.jpeg")
```
josemduarte, line 94: The `"_model-1.jpeg"` string should be extracted as a constant or perhaps even better a CLI parameter with default `"_model-1.jpeg"`

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,99 @@
...
+                fileObj = Path(outJpgFile)
+                if not (fileObj.is_file() and fileObj.stat().st_size > 0):
+                    logger.error("No image file: %s.", outJpgFile)
+            else:
+                logger.error('Missing bcif file %s', bcifFilePath)
```
josemduarte, line 99: This error should also not go silent for the workflow. An exception should be thrown

josemduarte (changes requested):
Thanks! Some comments from my side

piehld (changes requested):
Thanks for the updates @trumbullm. Just a handful of other small adjustments and suggestions below.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,97 @@
...
+                except subprocess.CalledProcessError:
+                    logger.exception("Failed to run cmd")
+
+                # check result
+                outJpgFile = os.path.join(outPath, fileId + checkFileAppend)
+                fileObj = Path(outJpgFile)
+                if not (fileObj.is_file() and fileObj.stat().st_size > 0):
+                    raise ValueError("No image file: ", outJpgFile)
+            else:
+                raise ValueError('Missing bcif file ', bcifFilePath)
```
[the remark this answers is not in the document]
trumbullm: Yeah I did this because the linter told me to. But I'll just do an f string instead.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -0,0 +1,97 @@
...
+                molrenderExe,
+                jpgRender,
+                bcifFilePath,
+                outPath,
+                "--height", jpgHeight,
+                "--width", jpgWidth,
+                "--format", jpgFormat,
+            ]
+            if jpgAdditionalCmds:
+                cmd = [*cmd, jpgAdditionalCmds]
```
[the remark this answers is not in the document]
trumbullm: @piehld Since it comes in from arguments I don't think it's possible to send a list. So I think this could stay assuming it's a string.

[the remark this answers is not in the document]
piehld: Hmm, OK...can you provide an example of what `jpgAdditionalCmds` might be? Can it be a string of many other commands? How many possible additional commands can there be? If it's only a few more, then I think it would make sense to create dedicated flags for each of them, since it seems kind of weird to me to hijack a single flag to pass in multiple commands.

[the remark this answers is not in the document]
josemduarte: I don't know of other parameters. In the current production workflow the parameters are in_file, out_path, height, width and format. You have all those covered with the other options. I'd say it is better to remove `jpgAdditionalCmds` if there's no use for it now.

trumbullm:
@josemduarte @piehld I removed the additional commands flag and tested it all works. It's ready to go if it has your two seals of approval.

josemduarte (approved):
LGTM thanks

piehld (approved):
Looks good to me too—thanks for all the edits!

PR #21, opened 2025-01-28
Correct installation instructions

dmyersturnbull:
This tells everyone to use `--use-deprecated=legacy-resolver` and `--extra-index-url https://pypi.anaconda.org/OpenEye/simple`.

Dennis, I wasn't quite sure about MySQL. I think it's always required, but maybe it's only needed when it falls back to compiling some packages. It doesn't appear documented elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,41 +2,58 @@
 
 [![Build Status](https://dev.azure.com/rcsb/RCSB%20PDB%20Python%20Projects/_apis/build/status/rcsb.py-rcsb_workflow?branchName=master)](https://dev.azure.com/rcsb/RCSB%20PDB%20Python%20Projects/_build/latest?definitionId=21&branchName=master)
 
-## Introduction
-
 RCSB Python workflow entry points for data processing and ETL/ELT operations.
 
-### Installation
+## Setup
 
-Download the library source software from the project repository:
+> [!IMPORTANT]
+> These instructions use [uv](https://docs.astral.sh/uv/).
+> To install it, run
+> ```bash
+> uv -V || curl -L -f https://astral.sh/uv/install.sh | sh
+> ```
+> Unfortunately, `rcsb.workflow` is not yet friendly to uv or modern pip:
+> To work around this, I recommend following the instructions below exactly.
+
+### To use the package
+
+Navigate to a directory you want to create a `.venv` in.
+Then run
 
 ```bash
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,41 +2,58 @@
...
-# or to make sure the submodules are updated --
-git submodule update --recursive --init
-git submodule update --recursive --remote
+> [!IMPORTANT]
+> These instructions use [uv](https://docs.astral.sh/uv/).
+> To install it, run
+> ```bash
+> uv -V || curl -L -f https://astral.sh/uv/install.sh | sh
+> ```
+> Unfortunately, `rcsb.workflow` is not yet friendly to uv or modern pip:
```
trumbullm, line 15: @dmyersturnbull what steps need to be taken to switch to a `pyproject.toml` file?

README.md
piehld, line 1: While I support the addition of the `uv` install instructions, I would greatly appreciate keeping the `pip` instructions around as well, since I personally haven't made the switch to `uv` yet. So maybe keep a section that is "Install with `pip`", and "Install with `uv`".

For the pip command, can it be updated to be:
`pip3 install --extra-index-url https://pypi.anaconda.org/OpenEye/simple "rcsb.workflow>=0.47"`

If we can avoid using the `--use-deprecated=legacy-resolver`, that would be preferable I think.

piehld (changes requested):
Thanks @dmyersturnbull, just have a few small requests.

README.md
dmyersturnbull: > Also, can you please keep the `git` commands around at the top of the file?
> 
> ```
> git clone  --recurse-submodules https://github.com/rcsb/py-rcsb_workflow.git
> 
> # or to make sure the submodules are updated --
> git submodule update --recursive --init
> git submodule update --recursive --remote
> ```

Up to you, but I recommend separating (A) wanting to develop or run tests (so cloning); and (B) just wanting to run commands (just pip/uv).

Also, the `git submodule` commands are unnecessary with `--recurse-submodules`. I always avoid listing too many options because it makes instructions longer and tends to be distracting.

That's also why I didn't show both uv and pip. Pip doesn't automatically make a venv, so `python -m venv` should probably be shown alongside.

README.md
dmyersturnbull: It's actually not too difficult. I started a branch, but I'll need to fix it up.

PR #22, opened 2025-01-29
ExDB(workflow, exec, and tests) and Dockerfile

trumbullm:
Adding ExDB(workflow, exec, and tests) and Dockerfile to stop circular imports from rcsb.exdb


piehld:
Thanks for all the PRs @trumbullm! Before you do anything more with them though, I'm actually having second thoughts on whether we should move the _entire_ CLI file over or just the parts that use `rcsb.workflow`. Let me think about it some more...and regardless I will for sure be able to help with making those changes.

trumbullm:
@piehld Sounds good! Just let me know once you decide. I might continue to poke around at this branch just because the tests keep failing and I want to understand why. Otherwise I'll just leave it here for now unless you want me to retract the merge request.

rcsb/workflow/wuw/ExDbWorkflow.py
```diff
@@ -0,0 +1,521 @@
...
+# File: ExDbWorkflow.py
+# Date: 17-Dec-2019  jdw
+#
+#  Workflow wrapper  --  exchange database loading utilities --
+#
+#  Updates:
+#   2-Mar-2023 dwp Add "numProc" parameter to 'upd_ref_seq' operation methods
+#   9-Mar-2023 dwp Lower refChunkSize to 10 (UniProt API having trouble streaming XML responses)
+#  25-Apr-2024 dwp Add arguments and methods to support CLI usage from weekly-update workflow
+#  20-Aug-2024 dwp Add LoadTargetCofactors step; change name of UpdateTargetsCofactors step to UpdateTargetsData
```
piehld, line 11: For this and other files `*.py`, can you please just add a comment noting that you moved these scripts from `rcsb.exdb`? (Appreciate what you've already noted in the HISTORY.txt!)

piehld (changes requested):
Thanks @trumbullm! This looks great. I ended up deciding just to stick with the original plan, since most of the code would have to move anyway. I just made a couple of minor comments. 

We should just coordinate again before you merge too, just to spot check things.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    && rm /opt/mmseqs-linux-avx2.tar.gz \
+    && ln -s /opt/mmseqs2/bin/mmseqs /usr/local/bin/mmseqs
+
+# Install the required Python packages
+RUN pip install --no-cache-dir --upgrade "pip>=23.0.0" "setuptools>=40.8.0" "wheel>=0.43.0" \
+    && pip install --no-cache-dir --user -r /app/requirements.txt \
+    && pip install --no-cache-dir "pymongo>=4.10.1"
+
+# Install the latest version of THIS packages
+RUN pip install --no-cache-dir "rcsb.exdb>=1.26"
```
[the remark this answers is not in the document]
josemduarte: One important consideration here: if this is to build an up-to-date image, then in the CI process we need to wait for the pip package to be published. Otherwise it will install the previous version. Is the CI workflow script written in such a way? That or make the docker installation based on the source code. 

ChatGPT gives these 2 ways:

1. FROM code in the repo: You copy your source code into the image directly (e.g., `COPY . /app` and then `pip install .`).
2. FROM PyPI: You create a lightweight Dockerfile that just installs the package from PyPI (e.g., `RUN pip install my-package==1.2.3`).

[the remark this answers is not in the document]
dmyersturnbull: @josemduarte is right, but it should definitely be `COPY . /app` and `pip install .` for exactly that reason. If the Dockerfile installs from PyPi, it can't be used for testing.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# Install mmseqs2
+ADD https://github.com/soedinglab/MMseqs2/releases/download/13-45111/mmseqs-linux-avx2.tar.gz /opt/
+RUN mkdir -p /opt/mmseqs2 \
+    && tar xzf /opt/mmseqs-linux-avx2.tar.gz -C /opt/mmseqs2 --strip-components=1 \
+    && rm /opt/mmseqs-linux-avx2.tar.gz \
+    && ln -s /opt/mmseqs2/bin/mmseqs /usr/local/bin/mmseqs
+
+# Install the required Python packages
+RUN pip install --no-cache-dir --upgrade "pip>=23.0.0" "setuptools>=40.8.0" "wheel>=0.43.0" \
+    && pip install --no-cache-dir --user -r /app/requirements.txt \
```
dmyersturnbull, line 29: It shouldn't be necessary to install the requirements separately; simply `pip install --no-cache-dir .` will work. Also, you'll unfortunately need `--extra-index-url https://pypi.anaconda.org/OpenEye/simple` for now.

dmyersturnbull: If you use pip, I recommend these lines:

```
ENV PYTHONFAULTHANDLER=1
ENV PYTHONUNBUFFERED=1
ENV PIP_NO_CACHE_DIR=on
ENV PIP_DISABLE_PIP_VERSION_CHECK=on
ENV PIP_DEFAULT_TIMEOUT=120
```

Dockerfile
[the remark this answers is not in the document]
dmyersturnbull: If you want to use uv instead, this might work:

```
ENV UV_NO_CACHE=1
ADD https://astral.sh/uv/install.sh /uv-installer.sh
RUN sh /uv-installer.sh && rm /uv-installer.sh

WORKDIR /app
COPY . /app/

RUN uv venv --python 3.10 --seed --no-cache
RUN uv pip install \
  --no-cache \
  --index https://pypi.anaconda.org/OpenEye/simple \
  .
```

After switching to pyproject, those two uv commands can be replaced with:

```
RUN uv sync --frozen --no-dev
```

piehld:
@dmyersturnbull can you explain why that is so important? The non-CLI code here is solely intended to be used by the CLI code, which is why they've been included in the same package. Is there a practical reason why they shouldn't? 

Even our `rcsb.db` package follows the same strategy, with its CLI existing in the same repo as the code it uses. 

Of course, most of this was done prior to my arrival, so I've been following suit. But if there's a clear reason to avoid this approach, then for sure we ought to reconsider how to structure each of our packages.

rcsb/workflow/tests/testExDbWorkflow.py
```diff
@@ -0,0 +1,144 @@
...
+        # self.__chemEtlD = {
+        #     "fetchLimit": 4,
+        #     "numProc": 1,
+        #     "chunkSize": 20,
+        #     "loadType": "full",
+        #     "ccUrlTarget": ccUrlTarget,
+        #     "birdUrlTarget": birdUrlTarget,
+        #     "ccFileNamePrefix": ccFileNamePrefix,
+        # }
+        #
```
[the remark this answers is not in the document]
trumbullm: It was actually tox that got mad at me which is interesting because it apparently didn't get mad in it's old repo. Is tox setup differently on rcsb.exdb?

rcsb/workflow/tests/testExDbWorkflow.py
```diff
@@ -0,0 +1,144 @@
...
+        # logger.info("Now running ExDbWorkflow for computed-models")
+        # try:
+        #     opL = ["upd_ref_seq_comp_models"]
+        #     rlWf = ExDbWorkflow(**self.__commonD)
+        #     for op in opL:
+        #         ok = rlWf.load(op, **self.__loadCommonD)
+        #         self.assertTrue(ok)
+        # except Exception as e:
+        #     logger.exception("Failing with %s", str(e))
+        #     self.fail()
```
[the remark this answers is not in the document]
trumbullm: I'll clear unused code out but it keeps failing the test. @piehld did you write these tests? Could you take a look at why it might be failing?

Dockerfile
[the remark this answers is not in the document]
trumbullm: I'll make some modification (I just copied this from .exdb) but I don't know how our testing infrastructure works for docker images. Does it go through tests after it's built?

pyproject.toml
```diff
@@ -0,0 +1,248 @@
...
+  "rcsb.utils.seq >= 0.82",
+  "rcsb.utils.targets >= 0.82",
+  "rcsb.utils.seqalign >= 0.31",
+  "rcsb.exdb >= 1.25",
+  "rcsb.utils.io >= 1.48",
+  #****** Miscellaneous dependencies ******#
+  "pymongo>=4.10.1", # from original dockerfile
+]
+
+[project.scripts]
```
trumbullm, line 71: @dmyersturnbull I'm having trouble adding clis outside of the uv env. Here it install it correctly but it only runs in uv. 
```
root@d76c968ebaf2:/app# imgs_exec_cli
bash: imgs_exec_cli: command not found
root@d76c968ebaf2:/app# uv run imgs_exec_cli
Installed 8 packages in 113ms
usage: imgs_exec_cli [-h] --op {genLists,genJpgs} [- ...
```

trumbullm: Nevermind. I added it to PATH `ENV PATH="/app/.venv/bin:$PATH"`

rcsb/workflow/tests/testExDbWorkflow.py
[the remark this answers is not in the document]
piehld: This was code that I had started writing but never finished, so left it commented out. We no longer need this now though, so you can delete it.

Let me look into the Azure failures...

piehld:
@trumbullm Just remembered, we'll also want to copy over the `.github/workflows` stuff from `rcsb.exdb` (modifying it as needed).

.github/workflows/workflow-docker.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  run-workflow:
+    name: "Run automated docker workflow"
+    uses: rcsb/devops-cicd-github-actions/.github/workflows/workflow-docker.yaml@master
+    with:
+      dockerfile_location: "Dockerfile"  # The location of the Dockerfile relative to the root of the repository. Defaults to "Dockerfile".
+      repo_project: "rcsb"  # REQUIRED. The name of the project or organization in the remote Docker image repository.
+      docker_image_name: "rcsb-exdb"  # REQUIRED. The name of the Docker image to create.
+      docker_build_context: "."  # The path location of the docker build context, relative to the project root. Defaults to the project root.
```
piehld, line 33: Think you need to change these for workflow

piehld: Yeah sorry, I highlighted more than necessary

Dockerfile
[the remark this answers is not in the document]
dmyersturnbull: > I'll make some modification (I just copied this from .exdb) but I don't know how our testing infrastructure works for docker images. Does it go through tests after it's built?

I don't think there's any automated testing. (Ref: [`azure-template-tox-job.yml`](https://github.com/rcsb/py-rcsb_workflow/blob/master/azure-template-tox-job.yml).)

> Since the Dockerfile is being rewritten we might as well move to `pyproject.toml` right? This seems like a good opportunity for it.

I'm workin' on it :) That's also a substantial change, so it might be good to separate the two anyway.

tox.ini
```diff
@@ -68,6 +74,8 @@ commands =
     {envpython} -m unittest discover -v --start-directory {[local_settings]test_path_2} --pattern "{[local_settings]test_pattern}"
     {envpython} -m unittest discover -v --start-directory {[local_settings]test_path_3} --pattern "{[local_settings]test_pattern}"
     {envpython} -m unittest discover -v --start-directory {[local_settings]test_path_4} --pattern "{[local_settings]test_pattern}"
```
piehld, line 76: I think you need to add the fixture paths to the first few commands instead of below them, e.g. see the order at https://github.com/rcsb/py-rcsb_exdb/blob/master/tox.ini#L77-L80

piehld (changes requested):

rcsb/workflow/tests/fixturePdbxLoader.py
```diff
@@ -0,0 +1,291 @@
...
+    #     fU = FileUtil()
+    #     modelSourcePath = os.path.join(self.__mockTopPath, "AF")
+    #     for iPath in glob.iglob(os.path.join(modelSourcePath, "*.cif.gz")):
+    #         fn = os.path.basename(iPath)
+    #         uId = fn.split("-")[1]
+    #         h3 = uId[-2:]
+    #         h2 = uId[-4:-2]
+    #         h1 = uId[-6:-4]
+    #         oPath = os.path.join(self.__cachePath, "computed-models", h1, h2, h3, fn)
+    #         fU.put(iPath, oPath)
```
piehld, line 211: Can you delete this?

rcsb/workflow/tests/fixturePdbxLoader.py
```diff
@@ -0,0 +1,291 @@
+##
+# File:    PdbxLoaderFixture.py
+# Author:  J. Westbrook
+# Date:    4-Sep-2019
+# Version: 0.001
+#
+# Updates:
+#
```
piehld, line 8: Can you add an update comment mentioning that this was copied from rcsb.exdb?

tox.ini
```diff
@@ -4,9 +4,15 @@
 # project specific local settings
 test_pattern = "test*.py"
 #
+fixture_pattern_1 = "fixtureD*.py"
```
piehld, line 7: I think you can drop this, since we didn't end up having to copy over the other fixture file.

tox.ini
```diff
@@ -64,6 +70,8 @@ deps =
 commands =
     echo "Starting {envname} with {envpython}"
     {envpython} -V
+    {envpython} -m unittest discover -v --start-directory {[local_settings]fixture_path_1} --pattern "{[local_settings]fixture_pattern_1}"
```
piehld, line 73: And drop this too

tox.ini
```diff
@@ -137,6 +145,8 @@ deps =
 commands =
     echo "Starting {envname}"
     coverage erase
+    coverage run --parallel-mode --omit="{[local_settings]coverage_exclude_paths}" --source="{[local_settings]coverage_source_paths}" -m unittest discover -v --start-directory {[local_settings]fixture_path_1} --pattern "{[local_settings]fixture_pattern_1}"
```
piehld, line 148: and this

tox.ini
```diff
@@ -4,9 +4,15 @@
 # project specific local settings
 test_pattern = "test*.py"
 #
+fixture_pattern_1 = "fixtureD*.py"
+fixture_pattern_2 = "fixtureP*.py"
+fixture_pattern_3 = ""
 # Source pathls (unquoted space separated list of files/directories) for linting and format checks
 source_paths = rcsb/workflow  setup.py
 #
+fixture_path_1 = "rcsb/workflow/tests"
```
piehld, line 13: and this

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    && pip install --no-cache-dir "pymongo>=4.10.1"
+
+# Install the local code
+WORKDIR /app  
+COPY . /app/
+RUN pip install .
```
piehld, line 35: We'll want to make sure with Jeremy that this install strategy will be sufficient for k8s/production weekly-update purposes

piehld (changes requested):
Thanks @trumbullm! Looks great overall. Just a few tidy-ness suggestions.

Regarding the cleanup of "setup.py", feel free to reassign "fixture..._2" to "fixture..._1" if you like. But not strictly necessary i don't think. Leaving it up to you.

Dockerfile
trumbullm: @jeremy-rcsb Could you take a look at this Dockerfile? We moved it over from rcsb.exdb to serve as the new image for workflow. What we changed was instead of [pulling from pypy](https://github.com/rcsb/py-rcsb_exdb/blob/master/Dockerfile#L33) we add the local repo and `pip install .`. Will this work with our current infrastructure?

jeremy-rcsb: Looks good to me. I don't see any reason why this wouldn't work.

josemduarte (approved):
LGTM thanks

PR #23, opened 2025-02-11
Mjt update dockerfile

trumbullm:
Removed secondary imgs dockerfile and added molrender (an additional ~30mb) to the workflow dockerfile.


.github/workflows/workflow-docker.yaml
```diff
@@ -21,6 +16,18 @@ on:
     #  │ │ │ │ │
     #  * * * * *
     - cron: "0 0 * * FRI"  # 0 UTC Friday
+  #
+  # Also run for new PRs and PR updates
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
```
josemduarte, line 25: With this you will get a build for any branch that is pull requested. I guess we can try that and see if it works for us. The alternative would be to build on push to any branch, which is useful when someone is developing and needs to test before pull-requesting.

josemduarte: In any case: one thing that is missing below is having an image name specific to the branch. I think we already have some sort of convention for that in Harbor, right @jeremy-rcsb ?

josemduarte (changes requested):
Thanks! See comment about image name.

.github/workflows/workflow-docker.yaml
trumbullm: @josemduarte As for building other branches I don't think we need to. I setup molrender in this Dockerfile to share. It only adds about 30mb more data to it which surprised me. Probably because it isn't adding bcif. So I don't think we need to build other branches any more.

josemduarte: If you strictly want no other branches, then you should remove the pull_request trigger. But, as I said, it is likely to be useful in the future for development (as long as the branch name is properly tagged)

piehld: Ah, that's a good point. I think we only want things merged to master to be pushed to harbor, not all successful workflow runs of any non-master branch.

Looking at our github workflow setup, is this already accounted for [here](https://github.com/rcsb/devops-cicd-github-actions/blob/master/.github/workflows/workflow-docker.yaml#L28-L31) (with default `master`)? I.e., only pushes to `master` will actually publish to harbor?

PR #24, opened 2025-02-17
RO-4458 new bcif workflow

j-s-135:
command line interface, test not implemented yet


epeisach:
I am not reviewing - but was just looking.  Seems to me that logging.basicConfig should be used in an app.  Support libraries should not declare logging levels, etc.  

A typical setup in libraries is to add

import logging
logger = logging.getLogger(__name__)

logger.error....

The application can then decide the logging level.  Otherwise - depending on import order - the application might have no control.

j-s-135:
Thanks, I will update that.

rcsb/workflow/bcif/bcif_functions.py
```diff
@@ -0,0 +1,44 @@
...
+   if python_molstar_java not in options:
+      logging.critical("error - language %s not yet supported" % python_molstar_java)
+      return False
+   if python_molstar_java == "molstar":
+      return molstar_convert(infile, outfile, molstar_cmd)
+   elif python_molstar_java == "python":
+      return py_convert(infile, outfile, workpath, da)
+
+def molstar_convert(infile:str, outfile:str, molstar_cmd:str) -> bool:
+   args = ["node", molstar_cmd, "-i", infile, "-ob", outfile]
```
trumbullm, line 20: If we are moving away from node molstar to pymmcif can we remove this?

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,367 @@
...
+    """ mounted paths must be already made """
+    if not os.path.exists(workflow_utility.updateBase):
+        os.makedirs(workflow_utility.updateBase, mode=0o777)
+    for dir in workflow_utility.contentTypeDir.values():
+        path = os.path.join(workflow_utility.updateBase, dir)
+        if not os.path.exists(path):
+            os.mkdir(path, mode=0o777)
+    return True
+
+def get_pdb_list_(workflow_utility: WorkflowUtilities, load_type: str, list_file_base: str, pdb_list_filename: str, result: bool) -> bool:
```
trumbullm, line 49: My understanding is this loads the text file and resaves it with pickle. Can we skip pickle all together?

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,367 @@
...
+    # save full tasks file
+    logging.info("get local tasks saving %d tasks to %s" % (len(tasks), input_list_2d))
+    with open(os.path.join(list_file_base, input_list_2d), "wb") as w:
+        pickle.dump(tasks, w)
+    # return list of task indices
+    tasks = list(range(0, len(tasks)))
+    logging.info("returning %d tasks" % len(tasks))
+    return tasks
+
+def split_list(nfiles:int, subtasks:int, tasklist:List[str]) -> List[List[str]]:
```
trumbullm, line 159: This split lists function might be able to be replaced with [splitIdList](https://github.com/rcsb/py-rcsb_db/blob/master/rcsb/db/wf/RepoLoadWorkflow.py#L283)

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,367 @@
...
+    start_file = os.path.join(list_file_base, status_start_file)
+    dirs = os.path.dirname(start_file)
+    if not os.path.exists(dirs):
+        os.makedirs(dirs, mode=0o777)
+    with open(start_file, "w") as w:
+        w.write("Binary cif run started at %s." % str(datetime.datetime.now()))
+    return True
+
+
+def branching_(r: int) -> str:
```
trumbullm, line 28: I think this routing should be done in the scheduler

trumbullm (changes requested):
It's looking good! I've commented some ideas for changes and some questions.

trumbullm:
Do we need to add `pymmcif` to our requirements.txt?

rcsb/workflow/bcif/bcif_functions.py
j-s-135: Yes, eventually.

rcsb/workflow/bcif/task_functions.py
j-s-135: We could since the data is just text, but then we'd have to convert the array to string.

rcsb/workflow/bcif/task_functions.py
j-s-135: Good point. I will move that to tasks.py in the other repository.

rcsb/workflow/bcif/bcif_functions.py
trumbullm: Would it be possible to do the transition within this merge? That way we could avoid adding Molstar to the Docker Image unnecessarily.

rcsb/workflow/bcif/task_functions.py
trumbullm: Would that be more efficient since it doesn't involve any extra read/writes?

rcsb/workflow/bcif/bcif_functions.py
j-s-135: All right.

rcsb/workflow/bcif/task_functions.py
j-s-135: If I switch to splitIdList then I will have to use text files. Otherwise, though, the strings would get converted back to an array anyway.

rcsb/workflow/bcif/workflow_functions.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def getWhatCoastWeAre(self) -> str:
+        coastPath = self.whatCoast
+        if coastPath is None or coastPath == "":
+            raise ValueError("The whatCoast parameter was not passed in luigi CLI.")
+        if coastPath not in ("east", "west"):
+            raise ValueError(
+                "The whatCoast parameter was not passed correctly in luigi CLI. It must be either 'east' or 'west'"
+            )
+        return coastPath
```
piehld, line 139: So these two functions are done at the `weekly-update-workflow` level, so you shouldn't need to add them here. I think what you would do here is just plan to accept the pre-formed prerelease filepath as a CLI argument. 

If you're trying to use these to make sure the user actually provided the necessary CLI arguments, I'd recommend just using `argparse` to do that, e.g.:

```python
    parser.add_argument(
        "--coast",
        default=None,
        required=True,
        ...
```

rcsb/workflow/bcif/workflow_functions.py
```diff
@@ -0,0 +1,242 @@
...
+    def getAllCurrentModelIdsWithMetadata(self) -> dict:
+        try:
+            # http://computed-models-internal-%s.rcsb.org/staging % east, holdings/computed-models-holdings.json.gz
+            holdingsFileUrl = os.path.join(
+                self.getCsmFileRepoBaseUrl(), self.csmHoldingsUrl
+            )
+            dic = {}
+            with urllib.request.urlopen(holdingsFileUrl) as url:
+                data = json.loads(gzip.decompress(url.read()))
+                for modelId in data:
```
piehld, line 200: Instead of using `urllib` and decompressing things here, you should be able to just use `MarshalUtil`, which does gzip expansion for you automatically and fetches web resources. E.g.,:

```python
from rcsb.utils.io.MarshalUtil import MarshalUtil
mU = MarshalUtil()
data = mU.doImport("https://files.wwpdb.org/pub/pdb/holdings/current_file_holdings.json.gz", fmt="json")
```

requirements.txt
piehld, line 1: FYI—I just pushed a few updates to the `master` branch recently that made changes to this file, so be aware you'll need to merge into your branch for this file. (see current version at https://github.com/rcsb/py-rcsb_workflow/blob/master/requirements.txt)

rcsb/workflow/bcif/bcif_functions.py
```diff
@@ -0,0 +1,18 @@
+import logging
+from mmcif.api.DictionaryApi import DictionaryApi
+from rcsb.utils.io.MarshalUtil import MarshalUtil
+
+logger = logging.getLogger(__name__)
+
+
+def bcifconvert(infile: str, outfile: str, workpath: str, da: DictionaryApi) -> bool:
```
piehld, line 8: Here too, can you rename `da` to `dictionaryApi`? Or something more specific?

piehld (changes requested):
Thank you @j-s-135! This looks excellent. I made a few comments, suggestions, and questions throughout, but overall very nice looking work. (Love the typing!)

By the way, @trumbullm and I just merged a couple things to the `master` branch recently, so there will be a few conflicts you'll need to address (mainly in the `requirements.txt` file). So just a heads up about that. Might be easier to just deal with the conflicts on GitHub (if you scroll to the bottom of your PR, there's a button to "Resolve conflicts").

rcsb/workflow/bcif/workflow_functions.py
```diff
@@ -0,0 +1,242 @@
...
+source: github.com/rcsb/weekly-update-workflow
+"""
+
+
+class ContentTypeEnum(Enum):
+    EXPERIMENTAL = "experimental"
+    COMPUTATIONAL = "computational"
+
+
+class WorkflowUtilities:
```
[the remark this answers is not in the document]
j-s-135: Ok.

rcsb/workflow/bcif/workflow_functions.py
```diff
@@ -0,0 +1,242 @@
...
+import datetime
+import os
+import logging
+from typing import Tuple, Dict
+
+logger = logging.getLogger(__name__)
+
+
+"""
+source: github.com/rcsb/weekly-update-workflow
```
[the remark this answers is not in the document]
j-s-135: Which file/class were you referring to? The file itself or the one it's referencing?

rcsb/workflow/bcif/bcif_functions.py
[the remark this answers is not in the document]
j-s-135: Ok.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,474 @@
...
+
+
+def batchTask(
+    tasks,
+    localInputsOrRemote,
+    updateBase,
+    compress,
+    tempPath,
+    workflowUtility,
+    da,
```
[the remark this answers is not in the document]
j-s-135: Ok.

rcsb/workflow/bcif/workflow_functions.py
```diff
@@ -0,0 +1,242 @@
...
+    COMPUTATIONAL = "computational"
+
+
+class WorkflowUtilities:
+
+    def __init__(
+        self,
+        coast: str = None,
+        interpolation: bool = True,
+        out: str = None,
```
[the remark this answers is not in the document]
j-s-135: Ok.

rcsb/workflow/bcif/bcif_functions.py
j-s-135: Ok.

rcsb/workflow/bcif/workflow_functions.py
j-s-135: I'm not sure what you mean about those already being done at weekly-update-workflow level. It seems like those workflow functions and parameters were based on the functions by the same names in the bcif module of the weekly update workflow, and what I wrote then follows what's in the weekly-update-workflow now. Although those vars are now command line parameters, something has to do string interpolation to combine them, and it's not a bad idea to validate the coast parameter. I agree that having format strings as parameters is odd, but that's verbatim from the weekly-update-workflow in sandbox_config.py. In what I have written, if the user passes in a format string, they set interpolation true and that sets appendCoastToFtpFileBasePath true. If instead they pass in a full path, they set interpolation false. I agree that there's room for improvement, but I'm not certain what to do at this point. It seems similar to the housekeeping functions. If we don't include them in the new bcif workflow, we have to make sure that someone writes it into Luigi.

piehld: Perhaps I'm missing something then. @josemduarte, can you comment on this? I was thinking these types of methods would continue to exist in the weekly-update code, but if not when we switch to AirFlow, then should we be including them here?

rcsb/workflow/bcif/workflow_functions.py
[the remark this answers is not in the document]
piehld: The file itself, sorry. So in the file header docstrings, it would be good to provide some more information about what the classes and functions in that file are used for.

rcsb/workflow/bcif/workflow_functions.py
josemduarte: I agree with @piehld 's comments. Let me explain a little bit of the design philosophy: 

- `weekly-update-workflow` or the new `rcsb-wuw` should contain a minimal amount of code, a few housekeeping things and that's it. The main thing they do is defining the tasks and the dependency graph. Anything specific to the environment (e.g. coast, type of cluster k8s, openstack, nersc, location of PDB file archive) should be in configuration parameters in those repos
- `py-rcsb_workflow` should contain CLI executables that are completely agnostic to the environment and should be possible to run independently of the workflow engine. They need to get all of the environment-dependent information from CLI parameters. So for example if we need to pass the location of the file archive, then we do: `--baseUrlPdbFileTree=http://prereleaseftp-west.rcsb.org`

rcsb/workflow/bcif/workflow_functions.py
```diff
@@ -0,0 +1,242 @@
...
+        return pdbList
+
+    def getAllCurrentPdbIdsWithTimestamps(self) -> Tuple[dict, str]:
+        timesDic = {}
+        # http://prereleaseftp-%s.rcsb.org/pdb % east, holdings/released_structures_last_modified_dates.json.gz
+        allPdbIdsUrl = os.path.join(
+            self.getPrereleaseFtpFileBaseUrl(), self.pdbIdsTimestampFilePath
+        )
+        with urllib.request.urlopen(allPdbIdsUrl) as url:
+            data = json.loads(gzip.decompress(url.read()))
```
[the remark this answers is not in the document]
josemduarte: I think the CLI that @piehld and @trumbullm wrote is ready to be used exactly for the purpose needed by the new BCIF workflow (just using other parameters). See https://github.com/rcsb/py-rcsb_db/pull/84 . 

Check how it would look like in rcsb-wuw: https://github.com/rcsb/rcsb-wuw/blob/30e1590b125e7937f83407f185a29baa1ff6518b/src/imgs/tasks.py#L28-L46 . In your case `targetFileDir` would be the directory where you store the BCIF files and `targetFileSuffix` would be `bcif.gz`

So both `getPdbList` and `getAllCurrentPdbIdsWithTimestamps` shouldn't be needed.

josemduarte (changes requested):
Thanks! I've added one comment that I think is important to check. We really want to reuse the already available `splitIdList` CLI

rcsb/workflow/bcif/workflow_functions.py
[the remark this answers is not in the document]
j-s-135: Ok.

rcsb/workflow/bcif/workflow_functions.py
j-s-135: Ok. I have removed the parameters coast and interpolation as well as those functions and the corresponding functions for csm.

rcsb/workflow/bcif/workflow_functions.py
j-s-135: Ok, and for another function that was similar.

j-s-135:
Is it all right to have airflow run the task functions without invoking the CLI (by directly importing task_functions.py), or should airflow run the tasks through the CLI? (this seems counterintuitive, since a CLI is presumably for humans who want to run some tasks from the command line)

trumbullm:
> Is it all right to have airflow run the task functions without invoking the CLI (by directly importing task_functions.py), or should airflow run the tasks through the CLI? (this seems counterintuitive, since a CLI is presumably for humans who want to run some tasks from the command line)

The Airflow scheduler only builds the command (and flags) then sends it to the appropriate image. The command could be `python3 run_bcif_converter --flags` but I think we should stick with making it into a cli for consistency.

rcsb/workflow/bcif/task_functions.py
j-s-135: Switched to text files.

rcsb/workflow/bcif/task_functions.py
j-s-135: Switched to splitIdList.

rcsb/workflow/bcif/workflow_functions.py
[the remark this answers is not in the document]
j-s-135: Ok.

piehld:
Yes we should stick with the CLI. CLIs can be equally used by humans and called by scripts/tasks.

trumbullm:
> So it will require an op for each task function? (otherwise, airflow would have no pipeline, just 1 task)

I'm not sure but here is what it looks like for image_wf. One task splits uses splitIds to make, say, ten files of pdb ids to do then ten tasks are given their own files to run through. Are you referring to the --op flag?

josemduarte:
> So it will require an op for each task function? (otherwise, airflow would have no pipeline, just 1 task)

For BCIF, the Airflow workflow should have 2 main tasks (that correspond to 2 CLIs in this repo):

- **Find ids** Find the id list that needs processing, dumping it to one or multiple files. In my understanding this CLI is already avaiable, so no need to create any new code for this.
- **Encode to BCIF** For a given input id list (one of the output files from **Find ids**), read PDBx/mmCIF files and output one BinaryCIF file per id

These would be wired in Airflow like this:
BCIF-workflow drawio
It is essential that Airflow takes care of the task splitting, i.e. the parallelisation is driven by Airflow. That's because it is in Airflow where we can choose where and how to execute the tasks (Kubernetes, Superfacility API, etc)

j-s-135:
The new splitIdList function is maybe not working for csms. I'm still testing it and have not switched to CLI yet. For now, I have written a separate function that could hopefully merge into that one. Please let me know whether I should do that or roll back to the previous version.

piehld:
> > So it will require an op for each task function? (otherwise, airflow would have no pipeline, just 1 task)
> 
> I'm not sure but here is what it looks like for image_wf. One task splits uses splitIds to make, say, ten files of pdb ids to do then ten tasks are given their own files to run through. Are you referring to the --op flag?

@j-s-135 if you're referring to the `--op` flag like Michael asked, yes you will need multiple available options for `--op` (meaning "operation"). For example, see the options here for ExDB loading: https://github.com/rcsb/py-rcsb_db/blob/master/rcsb/db/cli/RepoLoadExec.py#L53-L58. Each one of them triggers a different function to run.

j-s-135:
About the splitIdList function, sorry for the alarm, it seems to work if I use a different holdings file that I just found out about. For the timestamp function that checks modification times, it needs to know the output folder structure, so it should maybe be the same as production, which would presumably have a series of subfolders. At this time, it writes all files to the same folder.

rcsb/workflow/bcif/config.yml
```diff
@@ -0,0 +1,4 @@
+site_info_configuration:
```
josemduarte, line 1: These configurations should be in `rcsb-wuw` (and over there, we should eventually revisit how to externalise them, but that's another discussion).

josemduarte (changes requested):
Thanks @j-s-135 . I've had another pass at the code and left a few more comments. The main issue I see is that we really don't want to do any task handling in this layer. Tasks are going to be handled automagically by Airflow. In here all that we need is the single task: take ids, output bcif files.

Also one important stylistic comment: I feel like handling errors and exceptions by passing boolean values around is not following best practices. Exceptions were invented for that purpose, so we should embrace them: let them bubble up if there's no way to handle them, or catch them as late as possible to salvage specific situations. I do realise the boolean style was followed in much of the existing code that came from John, but I'd advise against continue using that style in new code.

requirements.txt
j-s-135: Ok.

rcsb/workflow/bcif/config.yml
j-s-135: Ok.

rcsb/workflow/cli/BcifExec.py
```diff
@@ -0,0 +1,189 @@
...
+    except Exception as e:
+        raise Exception(str(e)) from e
+
+
+if __name__ == "__main__":
+    t = time.time()
+    try:
+        main()
+    except Exception as e:
+        logger.exception(str(e))
```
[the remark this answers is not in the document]
j-s-135: Ok.

rcsb/workflow/cli/BcifExec.py
```diff
@@ -0,0 +1,189 @@
...
+    args = parser.parse_args()
+
+    try:
+        (BcifWorkflow(args))()
+    except RuntimeError as e:
+        raise Exception(str(e)) from e
+    except ValueError as e:
+        raise Exception(str(e)) from e
+    except Exception as e:
+        raise Exception(str(e)) from e
```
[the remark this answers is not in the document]
j-s-135: Ok.

rcsb/workflow/wuw/BcifWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+                self.logException("make task list from local failed")
+
+        elif self.batch is not None:
+            fileglobs = [
+                "%s/pdbx_core_ids-*.txt" % self.listFileBase,
+                "%s/pdbx_comp_model_core_ids-*.txt" % self.listFileBase,
+            ]
+            procs = []
+            # load balancing
+            for fileglob in fileglobs:
```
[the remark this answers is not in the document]
j-s-135: Ok.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,659 @@
...
+                if not os.path.exists(path):
+                    os.mkdir(path)
+                    os.chmod(path, 0o777)
+    except Exception as e:
+        logger.error(str(e))
+        return False
+    return True
+
+
+def splitRemoteTaskLists(
```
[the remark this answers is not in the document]
j-s-135: Completed. The bcif task has been modularized.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+    if not bcifFilePath:
+        raise ValueError("failed to form bcif file path")
+    if os.path.exists(bcifFilePath):
+        # earlier timestamp ... overwrite
+        try:
+            os.unlink(bcifFilePath)
+            if os.path.exists(bcifFilePath):
+                raise Exception("file %s not removed" % bcifFilePath)
+        except Exception as e:
+            logger.exception(str(e))
```
josemduarte, line 294: One can discuss if failing to remove a file is a critical error or not. I'm going to say it is important enough for the workflow to fail. So my recommendation here is to remove the try/catch block so that the exception is seen by the workflow engine.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+                        shutil.chown(dirs, "ubuntu", "ubuntu")
+                    except PermissionError as e:
+                        pass
+                with open(cifFilePath, "ab") as w:
+                    for chunk in r.raw.stream(1024, decode_content=False):
+                        if chunk:
+                            w.write(chunk)
+                try:
+                    shutil.chown(cifFilePath, "ubuntu", "ubuntu")
+                    os.chmod(cifFilePath, 0o777)
```
josemduarte, line 268: As mentioned in other comments: let's not set any permissions as part of this task. It is too specific to the environment.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+            logger.exception(str(e))
+            return
+        try:
+            if r and r.status_code < 400:
+                dirs = os.path.dirname(cifFilePath)
+                if not os.path.exists(dirs):
+                    os.makedirs(dirs)
+                    try:
+                        os.chmod(dirs, 0o777)
+                        shutil.chown(dirs, "ubuntu", "ubuntu")
```
josemduarte: As mentioned in other comments: let's not set any permissions as part of this task. It is too specific to the environment.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+        url = os.path.join(remotePath, pdbId[-3:-1], remoteFileName)
+        cifFilePath = os.path.join(tempPath, remoteFileName)
+        if contentType == "csm":
+            url = os.path.join(
+                remotePath, pdbId[0:2], pdbId[-6:-4], pdbId[-4:-2], remoteFileName
+            )
+            cifFilePath = os.path.join(tempPath, remoteFileName)
+        try:
+            r = requests.get(url, timeout=300, stream=True)
+        except requests.exceptions.ConnectTimeout as e:
```
josemduarte: Definitely this is a critical error. Let's not catch it and instead let it be seen by the workflow engine.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+            logger.exception(str(e))
+            return
+        try:
+            if r and r.status_code < 400:
```
josemduarte, line 255: I think here we want to be more strict:
```suggestion
            if r and r.status_code != 200:
```

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+                        if chunk:
+                            w.write(chunk)
+                try:
+                    shutil.chown(cifFilePath, "ubuntu", "ubuntu")
+                    os.chmod(cifFilePath, 0o777)
+                except PermissionError as e:
+                    pass
+            else:
+                raise requests.exceptions.RequestException(
+                    "error - request failed for %s" % url
```
josemduarte, line 273: Could you add the status code to the message? That will be useful for debugging

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+                )
+        except Exception as e:
```
josemduarte, line 275: I think we should not catch here either. Let the exception be seen by the workflow engine.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+        result = convert(cifFilePath, bcifFilePath, dtemp, dictionaryApi)
+        if not result:
+            raise Exception("failed to convert %s" % cifFilePath)
+        try:
+            shutil.chown(bcifFilePath, "ubuntu", "ubuntu")
+            os.chmod(bcifFilePath, 0o777)
+        except PermissionError as e:
+            pass
+        counter[0] += 1
+    except Exception as e:
```
josemduarte, line 315: Let's not catch this one either. Let it be seen by the workflow engine so that it notifies us and/or retries the task.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+    infile: str, outfile: str, workpath: str, dictionaryApi: DictionaryApi
+) -> bool:
+    mu = MarshalUtil(workPath=workpath)
+    data = mu.doImport(infile, fmt="mmcif")
+    try:
+        result = mu.doExport(outfile, data, fmt="bcif", dictionaryApi=dictionaryApi)
+        if not result:
+            raise Exception()
+    except Exception as e:
+        raise Exception("error during bcif conversion: %s" % str(e)) from e
```
josemduarte, line 385: This try/catch doesn't have much effect (just changing the message slightly). I'd rather remove it. It will improve readability.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,414 @@
...
+    infile: str, outfile: str, workpath: str, dictionaryApi: DictionaryApi
+) -> bool:
+    mu = MarshalUtil(workPath=workpath)
+    data = mu.doImport(infile, fmt="bcif")
+    try:
+        result = mu.doExport(outfile, data, fmt="mmcif", dictionaryApi=dictionaryApi)
+        if not result:
+            raise Exception()
+    except Exception as e:
+        raise Exception("error during bcif conversion: %s" % str(e)) from e
```
josemduarte, line 399: This try/catch doesn't have much effect (just changing the message slightly). I'd rather remove it. It will improve readability.

josemduarte (changes requested):
Thanks @j-s-135 for all the improvements. I think this is very close to what we want now. I've had another pass and add a few comments, mostly about exception handling. The philosophy should be that the process should fail with non-0 status when a critical error occurs. In most situations, you can achieve that easily by not catching exceptions. The workflow engine then sees the non-0 status and considers the task failed.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,370 @@
...
+                outputContentType,
+                outputHash,
+                contentType,
+                dictionaryApi,
+                temppath,
+                maxTempFiles,
+            )
+            procs.append(multiprocessing.Process(target=batchTask, args=args))
+        for p in procs:
+            p.start()
```
josemduarte, line 136: Here you seem to be spawning subprocesses with the different subtasks, right? I think it should rather be done all within python, i.e. execute `singleTask()` in parallel. The best way to do it is something like this [SO post](https://stackoverflow.com/a/20549090) recommends.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,370 @@
...
+        removeTempFiles(tempPath=temppath)
+        counter[0] = 0
+
+
+def getBcifFilePath(
+    pdbId, outfileSuffix, updateBase, contentType, outputContentType, outputHash
+):
+    bcifFileName = "%s%s" % (pdbId, outfileSuffix)
+    bcifFilePath = None
+    if contentType == "pdb":
```
josemduarte, line 286: Don't we have a content type enum somewhere already @piehld @trumbullm ? That would be the ideal why to handle this to have more strong typing.

josemduarte (changes requested):
Thank you again for all the new changes. I've had another look and I added a few other comments about things I didn't notice before. Especially important is that spawning of subprocesses should be avoided.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,305 @@
...
+from typing import Optional
+from mmcif.api.DictionaryApi import DictionaryApi
+from mmcif.io.IoAdapterPy import IoAdapterPy as IoAdapter
+from rcsb.utils.io.MarshalUtil import MarshalUtil
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+ContentTypeEnum = Enum(
```
josemduarte, line 31: @piehld would you know if we have this enum elsewhere so that we can reuse it?

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,305 @@
...
+    if not os.path.exists(dirs):
+        os.makedirs(dirs)
+
+    # convert to bcif
+    result = convert(cifFilePath, bcifFilePath, temppath, dictionaryApi)
+    if not result:
+        raise Exception("failed to convert %s" % cifFilePath)
+
+
+def getRemoteHash(
```
josemduarte, line 211: How about merging this function and the one below (`getLocalHash`) by making `remotePath` and `remoteFileName` optional parameters (defaulting to `None`). That is to avoid duplicated code logic for hashing.

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,305 @@
...
+) -> bool:
+    mu = MarshalUtil(workPath=workpath)
+    data = mu.doImport(infile, fmt="mmcif")
+    result = mu.doExport(outfile, data, fmt="bcif", dictionaryApi=dictionaryApi)
+    if not result:
+        return False
+    return True
+
+
+def deconvert(
```
josemduarte, line 297: This one isn't used anywhere. Do you plan to use it later?

josemduarte (changes requested):
Thanks again, it's looking good! I added some comments for minor improvements. Otherwise I'm happy with it. @piehld would you be able to have another look too?

rcsb/workflow/bcif/task_functions.py
j-s-135: I apologize that this pull request has been rather disorganized. Most people would finish their code before making a pull request, but instead I have been working through ideas during the pull request. In fact, that's what I was told to do. About the deconvert function, it's a request from a previous reviewer. I guess it's a feature being saved for later.

requirements.txt
```diff
@@ -6,4 +6,6 @@ rcsb.utils.targets >= 0.82
 rcsb.utils.seqalign >= 0.31
 rcsb.utils.io >= 1.48
 rcsb.exdb >= 1.28  # v1.28 is when circular dependency was removed
-rcsb.db >= 1.800
\ No newline at end of file
+rcsb.db >= 1.800
+requests
+mmcif
```
piehld, line 11: Please add version numbers here (else CI/CD will complain I think)

rcsb/workflow/bcif/task_functions.py
j-s-135: Ok.

piehld (changes requested):
Thanks @j-s-135! I made a couple of comments on specific files. In addition, I have the following more general comments:

1. You'll also need to update the version number in:
  - https://github.com/rcsb/py-rcsb_workflow/blob/master/rcsb/workflow/cli/__init__.py
  - and add a new HISTORY line to https://github.com/rcsb/py-rcsb_workflow/blob/master/HISTORY.txt

2. Also, can you move the test files to the [mock-data repository](https://github.com/rcsb/py-rcsb_mock-data)? @trumbullm should be able to help advise you on the process involved in doing so (since I'll be away), as well as help on updating the mock-data submodule here. In short, the steps are:
    1. Clone the mock-data repo to your computer and make your own branch
    2. Create a new directory containing the test files
    3. Create a PR to master and request @trumbullm for review
    4. Merge the PR
    5. Come back to this repo (`py-rcsb_workflow`) and update the submodule with 
    ```
    git submodule update --recursive --init
    git submodule update --recursive --remote
    ```
    6. Update your `testBcifWorkflow` file to make use of the `mock-data` submodule for accessing the test files

rcsb/workflow/bcif/task_functions.py
josemduarte: Oh yes... definitely that one is not reusable. Then let's keep it as it is. Thanks

rcsb/workflow/bcif/task_functions.py
```diff
@@ -0,0 +1,305 @@
...
+logger.setLevel(logging.INFO)
+
+
+ContentTypeEnum = Enum(
+    "ContentTypeEnum",
+    [("EXPERIMENTAL", "pdb"), ("COMPUTATIONAL", "csm"), ("INTEGRATIVE", "ihm")],
+)
+
+
+def convertCifFilesToBcif(
```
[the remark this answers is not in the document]
piehld: I see, thanks for explaining and apologies I didn't read through your comments before giving my review. I understand your point, and agree that we should definitely be `raise`ing errors when they occur and relying on that for error handling. So that stuff should definitely stay in there.

But, I do see a benefit of returning a boolean (well, specifically `return True` at the the very end of the function), which is to provide a way of logging when the actual task is done. For example, you would run that function with the likes of:

```python
ok = convertCifFilesToBcif(...)
logger.info("Conversion completed with status %r", ok)
```

Then, when the function is called, you have a way of logging something explicit about whether it ran to completion or not. It may be redundant to just logging `"Conversion completed."` after calling it, but if there's a case where the function can exit successfully before it reaches the end, then it would be valuable to log either `"Conversion completed with status None"` (which would be the case if it exits early) vs `"Conversion completed with status True"` (which would be the case if it ran to the very end `return True`).

So, to clarify and adjust my original suggestion, what I suggest now is to simply put a `return True` at the very end of the function and adjust the docstrings, as follows (with help from GPT :) ):

```python
def convertCifFilesToBcif(...) -> Optional[bool]:
    """
    Converts CIF files to BCIF format based on a given list file.

    Returns:
        bool: True if the process completes successfully.
        None: If an error occurs or no explicit success state is provided.
    
    Raises:
        FileNotFoundError: If the specified list file does not exist.
        ValueError: If the list file is empty.
        RuntimeError: If processing fails for some files.
    """
    ...
    try:
        if os.path.exists(temppath):
            shutil.rmtree(temppath)
    except Exception as e:
        logger.error(str(e))

    return True
```

This makes it clear to users that:
1. The function _usually returns_ `True` if everything succeeds.
2. It _may return_ `None` if no explicit success condition is set.
3. It can raise exceptions for errors instead of returning a failure indicator.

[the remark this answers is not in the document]
josemduarte: @piehld I'd argue that can be achieved with exceptions. A calling example would look like this:
```python
try:
  convertCifFilesToBcif(my_cif_file)
  logger.info("Done converting %s succesfully" % my_cif_file)
catch Exception as e:
  logger.error(str(e))
  raise e
```
Note that if it does throw an exception the info logging message is never printed.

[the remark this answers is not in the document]
josemduarte: > That would work, but it depends on `convertCifFilesToBcif` properly raising an exception in every possible scenario. 

That is exactly what python should be doing: exceptions are what happens in any exceptional situation. I can see scenarios where some `None` values are not handled correctly and python is sort of lenient about it not throwing exceptions. In any case, that would be a bug in our code that needs fixing.

> With my proposal, there would be three possible outputs: `True`, `None`, or `Exception`. But with yours there are only two: `None` or `Exception`, so we wouldn't be able to differentiate between scenarios when the function runs to completion or exits early with `None`.

Wouldn't this alternative achieve the same as you propose without a return?
```python
def convertCifFilesToBcif(...) -> Optional[bool]:
    """
    Converts CIF files to BCIF format based on a given list file.

    Raises:
        FileNotFoundError: If the specified list file does not exist.
        ValueError: If the list file is empty.
        RuntimeError: If processing fails for some files.
    """
    ...
    try:
        if os.path.exists(temppath):
            shutil.rmtree(temppath)
    except Exception as e:
        logger.error(str(e))

    logger.info("Done cif to bcif conversion")
```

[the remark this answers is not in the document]
piehld: > That is exactly what python should be doing: exceptions are what happens in any exceptional situation. I can see scenarios where some None values are not handled correctly and python is sort of lenient about it not throwing exceptions. In any case, that would be a bug in our code that needs fixing.

I guess I was thinking more along the lines of if a loop is broken out of without error for some reason (e.g., an unintentional `break`), in which the function could exit successfully. I'm just thinking very generally, not about this particular function.

> Wouldn't this alternative achieve the same as you propose without a return?

Yeah that would achieve the same thing too. 

I don't have a super strong feeling either way, I guess I'm just used to always returning _something_, so that's why I made my original proposal. But I see your point and appreciate knowing the better code practice to follow.

piehld:
FYI @j-s-135, if you need the new mock-data available for tests in this repository, you'll need to update the submodule via:

```
# assuming you've already cloned the repo with recursive submodules:
git submodule update --recursive --remote

# if not, then also run:
git submodule update --recursive --init
git submodule update --recursive --remote
```

j-s-135:
It looks like the failures were caused by another file in the tests folder, fixturePdbxLoader.

rcsb/workflow/bcif/task_functions.py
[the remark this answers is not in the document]
piehld: @j-s-135 Apologies that our guidance was differing on this, but for this can you please follow @josemduarte's recommendation? I.e., do not return Bool for this function, and add logging at the end of this method instead, as shown below (along with updated docstrings):

```python
def convertCifFilesToBcif(...):
    """
    Converts CIF files to BCIF format based on a given list file.

    Raises:
        FileNotFoundError: If the specified list file does not exist.
        ValueError: If the list file is empty.
        RuntimeError: If processing fails for some files.
    """
    ...
    try:
        if os.path.exists(temppath):
            shutil.rmtree(temppath)
    except Exception as e:
        logger.error(str(e))

    logger.info("Done with CIF to BCIF conversion")
```

rcsb/workflow/wuw/BcifWorkflow.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __call__(self):
+
+        logger.info(
+            "running bcif workflow on %s and %s",
+            os.path.join(self.listFileBase, self.listFileName),
+            self.remotePath,
+        )
+
+        ok = convertCifFilesToBcif(
```
piehld, line 76: And because of that last requested change, you can avoid capturing "ok" and logging it here.

HISTORY.txt
```diff
@@ -46,3 +46,4 @@
                     Remove IMGT from feature building after service became unavailable February 2025
 19-Feb-2025 - V0.51 Bring back IMGT
  4-Apr-2025 - V0.52 Add multi cpu execution to jpg generation
+10-Apr-2025 - V0.53 Bcif cli
```
piehld, line 49:
```suggestion
10-Apr-2025 - V0.53 Add BCIF conversion CLI
```

rcsb/workflow/cli/BcifExec.py
```diff
@@ -0,0 +1,149 @@
...
+from rcsb.workflow.wuw.BcifWorkflow import BcifWorkflow
+
+logger = logging.getLogger("bcif")
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler()
+formatter = logging.Formatter(
+    fmt="%(asctime)s @%(process)s [%(levelname)s]-%(module)s: %(message)s"
+)
+handler.setFormatter(formatter)
+logger.addHandler(handler)
```
piehld, line 28: Can you move this logging setup to be within the `main()` function, following the argument parsing stuff? This would allow you to also add flags for `--debug` and `--log_file_path`, and would only be executed when the CLI is called, which is I think what we want to do. 

You could follow this as an example: https://github.com/rcsb/py-rcsb_db/blob/master/rcsb/db/cli/RepoLoadExec.py#L180-L199

And note how the `logging.getLogger()` is done at the top of the file: https://github.com/rcsb/py-rcsb_db/blob/master/rcsb/db/cli/RepoLoadExec.py#L48

rcsb/workflow/cli/BcifExec.py
```diff
@@ -0,0 +1,149 @@
...
+__author__ = "James Smith"
+__email__ = "james.smith@rcsb.org"
+__license__ = "Apache 2.0"
+
+import argparse
+import logging
+import time
+from rcsb.workflow.wuw.BcifWorkflow import BcifWorkflow
+
+logger = logging.getLogger("bcif")
```
piehld, line 21: Also, was the use of "bcif" intentional, or would `logging.getLogger(__name__)` work too?

rcsb/workflow/wuw/BcifWorkflow.py
```diff
@@ -0,0 +1,93 @@
...
+from rcsb.workflow.bcif.task_functions import convertCifFilesToBcif
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler()
+formatter = logging.Formatter(
+    fmt="%(asctime)s @%(process)s [%(levelname)s]-%(module)s: %(message)s"
+)
+handler.setFormatter(formatter)
+logger.addHandler(handler)
```
piehld, line 27: After doing the above, I think here you want to _remove_ the logging setup stuff, since that would have already been setup by the `BcifExec`. So, here, the only thing you'd keep is:

```python
logger = logging.getLogger(__name__)
```

rcsb/workflow/tests/testBcifWorkflow.py
```diff
@@ -0,0 +1,698 @@
...
+    deconvert,
+    getDictionaryApi,
+)
+
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s]-%(module)s.%(funcName)s: %(message)s",
+)
+
```
piehld, line 36: Here, for this `testBcifWorkflow` file, I think you want to _keep_ this, but should adjust it to as follows:

```suggestion
logging.basicConfig(
    level=logging.INFO,
    format="%(asctime)s [%(levelname)s]-%(module)s.%(funcName)s: %(message)s",
)
logger = logging.getLogger()
logger.setLevel(logging.INFO)
```

Then for all calls to `logging.info` or `.warning` etc. below, you'd change them to `logger.info` etc.

piehld (changes requested):
@j-s-135 Overall very close to finished, just a few comments based on the discussion Jose and I had on function output, and some notes on logging.

By the way, I resolved the error you were observing with Azure—it was due to a mongo installation problem (the commands to install it were outdated). I've merged that to master now, so please be sure to merge the latest master into your branch again and that should fix things.

rcsb/workflow/cli/BcifExec.py
j-s-135: It seemed like __name__ would get the parent logger and then adding a handler and formatter would overwrite those in the parent logger. However, it's working better now without adding a new name.

piehld (approved):
Thanks for all the changes, @j-s-135! Looks great to me. As long as the tests pass (which it looks like they are for your code), I say good to go!

josemduarte (approved):
LGTM thanks!

PR #27, opened 2025-02-24
adding xauth

trumbullm:
Small patch to add xauth. Needed for jpg generation


josemduarte (approved):
LGTM thanks

trumbullm:
@josemduarte tests are failing because a server is down.
> [ERROR]-FileUtil.__fetchUrlReqNoRt: Fetch 'https://card.mcmaster.ca/latest/ontology' fails with status 500
Is it okay if I merge anyways? Since I didn't change any of that code.

piehld:
Yeah this should be fine. But would be good to re-run the tests manually on Thursday just to make CARD is live again (it was down for sometime last Friday, which makes me worried).

PR #28, opened 2025-04-03
Added parallel execution in single runtime

trumbullm:
JpgGen can now run --numProcs number of parallel executions. This is helpful in containerized environments in both k8s and Nersc to reduce number of containers spun up.


trumbullm:
@piehld I don't know if this merits a version change.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -17,13 +17,34 @@
...
 
 from rcsb.utils.io.MarshalUtil import MarshalUtil
 
 logging.basicConfig(level=logging.INFO, format="%(asctime)s [%(levelname)s]-%(module)s.%(funcName)s: %(message)s")
 logger = logging.getLogger()
 
+# For pickling reasons this command needs to not accept self
+def run_command(args):
+    """Run a command and verify the output file."""
+    cmd, outPath, name, checkFileAppend = args
```
piehld, line 31: I think you need to put this within the class of `PdbCsmImageWorkflow` for it to accept `self`. Not sure if you intentionally wanted this outside the scope of the class or not.

piehld (changes requested):
Thanks @trumbullm. Left one comment for now. Also looks like some tests are failing on azure, so you should take a look a those.

> @piehld I don't know if this merits a version change.

Yes we will definitely want t increment the version, since this changes the actual code.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
trumbullm: It was intentional. Whenever I try to include this code in the class with `self` when it sends it to the `ProcessPoolExecutor` it gives a pickling error. As far as I can tell it is solved by only passing a single argument to it, so no `self`. Is having this in-class important? I can try to find a different work around if we need it all in class.

piehld: I guess I'm confused as well why it can't accept two arguments. To clarify, is what you tried the following:

```python
# Within the PdbCsmImageWorkflow class:
def run_command(self, args):
...

And then when calling it you would use:
self.run_command(args)
```

Are you saying that doesn't work?

If so, why not just define the function without `self`? It doesn't look like you use self anyway, so you could just do:
```python
def run_command(args):

```

Also to clarify, I don't think there is super important reason to keep these methods together in one class, but it would be cleaner to do so in my opinion, if possible.

trumbullm: I agree with the code clarity. At the very least I'll integrate it with the class. As for the `self` argument it refused with weird pickling errors but let me retry some other things and get back to you.

trumbullm: Okay it works with `self`. I think what happened I used a different mapping command previously which resulted in the pickle issue. But It works with this method. So let me know what you think

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -73,33 +77,50 @@ def imagesGenJpgs(self, **kwargs: dict) -> None:
...
+            logger.error("Stderr: %s", e.stderr)
+            raise
+
+        # Verify output file
+        outJpgFile = os.path.join(outPath, name + checkFileAppend)
+        outFileObj = Path(outJpgFile)
+        if not (outFileObj.is_file() and outFileObj.stat().st_size > 0):
+            raise ValueError(f"No image file generated: {outJpgFile}")
+
+        return f"Success: {cmd}"
```
piehld, line 126: Just curious, is it also feasible to just log the info message here, rather than iterating over `for result in results` above? Or does the `executor.map(...)` thing require you to do that?

E.g., could you just do here:
```python
logger.info("Success: %r", cmd)
# without a return statement
```
And then remove this part above:
```python
            for result in results:
                logger.info(result)
```

piehld (commented):
Thanks @trumbullm! Just two follow up comments

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -73,33 +77,50 @@ def imagesGenJpgs(self, **kwargs: dict) -> None:
...
+        else:
+            # Execute commands in parallel with process-based execution
+            with ProcessPoolExecutor(max_workers=int(numProcs)) as executor:
+                results = executor.map(self.run_command, args)
+            # Print results
+            for result in results:
+                logger.info(result)
+
+    # For pickling reasons this command needs to not accept self
+    def run_command(self, args):
```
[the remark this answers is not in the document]
trumbullm: Nevermind that. I have to use `self` which in turn auto passes in the self argument. When I make it a sub func I got 
> Subprocess failed: Can't pickle local object 'PdbCsmImageWorkflow.imagesGenJpgs.<locals>.run_command'

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -73,33 +76,50 @@ def imagesGenJpgs(self, **kwargs: dict) -> None:
...
+        else:
+            with ProcessPoolExecutor(max_workers=int(numProcs)) as executor:
+                futures = [executor.submit(self.run_command, arg) for arg in args]
+                for future in as_completed(futures):
+                    try:
+                        future.result()
+                    except Exception as e:
+                        logger.error("Subprocess failed: %s", str(e))
+
+    def run_command(self, args):
```
piehld, line 108: Oh right, in order to remove `self`, I think you need to make this a `staticmethod`, i.e.:
```python
    @staticmethod
    def run_command(args):
```
I'll leave it up to you whether you want to do that or just leave `self` here. I was just proposing this because of the limitations you mentioned earlier where the function needed exactly one argument.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -73,33 +76,50 @@ def imagesGenJpgs(self, **kwargs: dict) -> None:
...
+                    molrenderExe,
+                    jpgRender,
+                    bcifFilePath,
+                    outPath,
+                    "--height", jpgHeight,
+                    "--width", jpgWidth,
+                    "--format", jpgFormat,
+                ]
+                # passing in args this way is due to a pickling issue with multiple subprocesses
+                args.append((cmd, outPath, name, checkFileAppend))
```
piehld, line 94: Can this be called `argsL`? That's a little more clear that it is a list of tuples of args, versus a flattened list of args.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -73,33 +76,50 @@ def imagesGenJpgs(self, **kwargs: dict) -> None:
...
             else:
                 raise ValueError(f"Missing bcif file {bcifFilePath}")
+        if numProcs == 1:
+            self.run_command(args[0])
```
piehld, line 98: Are you intentionally only running the first args Tuple here? Even if `numProcs` is 1, wouldn't you still want to run through all the tuples of `args`, just one at a time?

Would this be more appropriate (assuming you've changed `args` to `argsL`?:
```python
        if numProcs == 1:
            for args in argsL:
                self.run_command(args)
```

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -73,33 +76,50 @@ def imagesGenJpgs(self, **kwargs: dict) -> None:
...
+        else:
+            with ProcessPoolExecutor(max_workers=int(numProcs)) as executor:
+                futures = [executor.submit(self.run_command, arg) for arg in args]
```
piehld, line 101: Assuming you change `args` above to `argsL`, can you adjust this to be:

```suggestion
                futures = [executor.submit(self.run_command, args) for args in argsL]
```

piehld (changes requested):
Thanks @trumbullm, sorry for the confusion earlier, I forgot to include the `@staticmethod` in my suggestion. But will leave the final decision to you on that.

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
trumbullm: Yes good catch

trumbullm:
Hey @piehld I needed to tack on this unzip and then delete command for this workflow. Jeremy and Jose both agree it's the best course of action right now to do it within the jpg gen pipeline. When Molrender allows for a .gz file then these two commands will be removed.

piehld (approved):
Thanks @trumbullm, I think this looks good!

PR #33, opened 2025-06-04
Dev jpggen subdirs for input and output

trumbullm:
- Requires hash dir for input `.bcif.gz` files. Now it always uses it.
- Added hash dir for output jpg files.
- Added common module for hashing function in anticipation of them getting more complicated.


trumbullm:
@josemduarte We discussed a common hashing function in anticipation for more complex id names. Here is my first attempt at it.

rcsb/workflow/hashUtils/idHash.py
piehld, line 1: My only comment right now is that we probably don't need an entire dedicated directory for this. I'd recommend just creating a file:`rcsb/workflow/wuw/WuwUtils.py`, in which you have this method here. Then, we could use that file to add other utility methods that we want in the future.

Also, note that your function here is written `idhash()` instead of `idHash()`. Not sure if that was intentional or not.

rcsb/workflow/cli/ImgExec.py
```diff
@@ -43,7 +43,6 @@ def main() -> None:
     parser.add_argument("--jpgRender", default="all", help="Which elements of the potein do you want to render.")
     parser.add_argument("--checkFileAppend", default="_model-1.jpeg", help="What jpg file do you want to check exists after the process runs.")
     parser.add_argument("--contentTypeDir", default="pdb", help="Is this list for pdb structures or csm?")
-    parser.add_argument("--useIdSubdir", action="store_true", help="Is the bcif file in a subdir using it's own name?")
```
piehld, line 46: So the plan is that this won't ever need to be used, and so can be removed?

rcsb/workflow/wuw/WuwUtils.py
piehld, line 1: Can you make sure there is a blank line at the end of this file? Sometimes GitHub doesn't display it, but often that single blank line is required.

rcsb/workflow/wuw/WuwUtils.py
```diff
@@ -0,0 +1,21 @@
+"""Hashing functions for file storage."""
+##
+# File: hashDirectories.py
```
piehld, line 3:
```suggestion
# File: WuwUtils.py
```

rcsb/workflow/wuw/WuwUtils.py
```diff
@@ -0,0 +1,21 @@
...
+#
+##
+__docformat__ = "google en"
+__author__ = "Michael Trumbull"
+__email__ = "michael.trumbull@rcsb.org"
+__license__ = "Apache 2.0"
+
+
+def idHash(name):
+    if "_" in name:
```
piehld, line 19: Is it possible to provide another argument, like "modelType" which would be "experimental" or "computational"? Or would that create too much headache? 

At the very least, I'd feel better if the `if "_" in name` was a little more explicit, e.g.:
```python
if name.lower().startswith("af_") or name.lower().startswith("ma_"):
```

rcsb/workflow/wuw/WuwUtils.py
```diff
@@ -0,0 +1,21 @@
...
+        return f"{name[0:2]}/{name[-6:-4]}/{name[-4:-2]}/"
+    return f"{name[-3:-1]}/"
```
piehld, line 21: Likewise, I think it would be prudent to check here that the name is only 4-charcters, and then add another "else" conditional raising an error if an ID doesn't fit into either of the above conditionals.

rcsb/workflow/wuw/WuwUtils.py
```diff
@@ -0,0 +1,21 @@
...
+#  04-Jun-2025 mjt Created hashing function
+#
+##
+__docformat__ = "google en"
+__author__ = "Michael Trumbull"
+__email__ = "michael.trumbull@rcsb.org"
+__license__ = "Apache 2.0"
+
+
+def idHash(name):
```
piehld, line 18: Can you change `name` to `id`? Just makes more sense to me.

Also, can you add a short docstring here with an example or two? (Feel free to use chatGPT to generate it :)

rcsb/workflow/wuw/PdbCsmImageWorkflow.py
```diff
@@ -60,20 +60,19 @@ def imagesGenJpgs(self, **kwargs: dict) -> None:
...
+            bcifFileName = os.path.join(idHash(name), name) + ".bcif.gz"
+            logger.info("%s running %s %s %s", i, name, bcifFileName, contentTypeDir)
 
             if contentTypeDir == "pdb":
                 bcifFilePath = os.path.join(pdbBaseDir, bcifFileName)
             else:
                 bcifFilePath = os.path.join(csmBaseDir, bcifFileName)
 
-            outPath = os.path.join(jpgsOutDir, contentTypeDir)
+            outPath = os.path.join(jpgsOutDir, contentTypeDir, idHash(name), name)
```
piehld, line 75: Rather than calling `idHash(name)` twice, I'd suggest calling it once above and assign it to a variable (e.g., `nameHash`), then just referencing that in the `bcifFileName = ...` and `outputPath = ...` lines.

piehld (changes requested):
Thanks @trumbullm! I just had a handful of very minor comments/suggestions, which I hope are easy enough to make. Let me know if you have any questions about them.

rcsb/workflow/cli/ImgExec.py
trumbullm: yes i think it already is removed here.

rcsb/workflow/wuw/WuwUtils.py
trumbullm: just checked. it's there

rcsb/workflow/wuw/WuwUtils.py
trumbullm: > Or would that create too much headache?
no but I like the logic being in this function instead of earlier. then it would have to be repeated. I'll add your more explicit check though.

rcsb/workflow/wuw/WuwUtils.py
piehld: Yes I just confirmed with Jasmine as well that that is the plan. However, the parent and subdirectory structure of the new archive will be different, so we will likely need to implement a new condition here to handle those uniquely when the time comes. It might be prudent for now to add something like:
```python
if name.lower().startswith("pdb_"):
    raise ValueError("Hashing method not implemented yet for extended PDB identifiers - %s" % name)
```

rcsb/workflow/wuw/WuwUtils.py
```diff
@@ -0,0 +1,23 @@
...
+__email__ = "michael.trumbull@rcsb.org"
+__license__ = "Apache 2.0"
+
+
+def idHash(structure_id):
+    if structure_id.lower().startswith("af_") or structure_id.lower().startswith("ma_"):
+        return f"{structure_id[0:2]}/{structure_id[-6:-4]}/{structure_id[-4:-2]}/"
+    if structure_id.lower().startswith("pdb_") or len(structure_id) == 4:
+        return f"{structure_id[-3:-1]}/"
+    raise RuntimeError("Unsupported structure id.")
```
piehld, line 23:
```suggestion
    raise RuntimeError("Unsupported structure id %s" % structure_id)
```

piehld (commented):
LGTM! Just had one more small suggestion. Assuming the Azure tests are passing, feel free to merge it after making that change.

PR #36, opened 2025-07-25
dev-lowercase-outputs-for-bcif

j-s-135:
Output files will be lower-case, including csm ids, while input csm ids will continue to be written in upper-case.


piehld:
Hi @j-s-135 @trumbullm, I worry this might affect production. Can you please hold off on merging this until I have a chance to review?

PR #38, opened 2025-09-04
update requirements.txt

j-s-135:
Bump rcsb db version to trigger rebuild of workflow image.


piehld:
@j-s-135 The workflow image will be built automatically this evening (based on its cron schedule), so this change isn't necessary. If you need to re-build the image immediately, I'd suggest just doing so manually via the GitHub actions UI (e.g., just re-run the last successful action).

(By the way, if we did need to do this, we'd need to also bump the version number of this package in `__init__.py`, since the requirements file is tightly coupled to the version of the published package.)

PR #50, opened 2026-07-21
new bcif hash

trumbullm:


trumbullm:
Tested with regular ids and ext ids. Works.

piehld (approved):
Thanks @trumbullm, LGTM! But, if you think of any other small tweaks that need to happen, please try to incorporate them into this PR/version bump if possible, so that we don't end up releasing too many different versions for incremental changes.

